PR #4, opened 2023-12-13
Skip failure on setting inactive interface as bond's primary

vcandapp:
The ovs command to set the primary interface of the bond fails when the interface is not active. This patch ignores the error and proceeds with the remaining configuration.

Change-Id: I3592af4f357dca483fc7914133e1fc248d12f11a (cherry picked from commit b4779a6b28e01689025919548ee3ef51c13c8188) (cherry picked from commit 01d1c57596aa0e61f04fbd4008cc43505f06246c)


karthiksundaravel (approved):
LGTM. THANKS

PR #5, opened 2023-12-14
VLANs needs to be restarted when the associated linux bond is restarted

vcandapp:
Consider the case where 2 interfaces eno1 and eno2 are added as members to linux bond bond1. And vlan100 is created on bond1.  Now when os-net-config is re-run it finds a change in ifcfg-eno1 or ifcfg-eno2. In this case the affected devices would be restarted, which triggers a restart of bond1. In this case the vlan devices associated with the bond shall also be restarted.

Change-Id: If80544262bf3827e5af0c32abcdcc5b6c2e71905 (cherry picked from commit ba37f85889a949b953c53c1085439c49028e3487)


karthiksundaravel (approved):
LGTM. Thanks

PR #6, opened 2023-12-20
Fix add_ovs_interface log which mislead ovs dpdk port

SharonBX:
Change-Id: Ib6678112bd9b3c4db4c97d517635d64121460cab


karthiksundaravel (approved):
LGTM !!!

PR #11, opened 2024-02-22
execute "ip link set dev <dev> down" if device is still active instead of "ifdown <interface>"

minmzzhang:
In NCS, we have a feature to add/remove vlan interface/s post install. To activate/deactivate the vlan interface/s, os-net-config was called upon such operation, and os-net-config invokes Linux ifdown command when restarting an interface. Linux ifdown-eth script deletes vlan type of interface/s.

The host level impact is vlan traffic loss. The impact on ipvlan pods is decremental as the ipvlan interfaces in the pods are deleted. Jing Zhang has made changes to os-net-config scripts to avoid vlan deletion when os-net-config is called.

### DIFF
--- a/os_net_config/__init__.py
+++ b/os_net_config/__init__.py
@@ -329,9 +329,10 @@ class NetConfig(object):
         if not self.noop:
             os.remove(filename)
 
-    def ifdown(self, interface, iftype='interface'):
-        msg = 'running ifdown on %s: %s' % (iftype, interface)
-        self.execute(msg, '/sbin/ifdown', interface, check_exit_code=False)
+    def ifdown(self, interface, iftype='interface', cleanup=True):
+        if cleanup:
+            msg = 'running ifdown on %s: %s' % (iftype, interface)
+            self.execute(msg, '/sbin/ifdown', interface, check_exit_code=False)
         if utils.is_active_nic(interface):
             msg = '%s %s is up, trying with ip command' % (iftype, interface)
             self.execute(msg, '/sbin/ip',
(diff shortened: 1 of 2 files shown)


karthiksundaravel:
The restart of the interfaces are required typically when there is a change in the ifcfgs. Can you please help me understand if there is any change in ifcfgs. There are certain changes like MTU, Ethtool opts, ip addr that doesn't require reboot.

karthiksundaravel:
Can you please share the config.yaml before and after the change

dsneddon:
> we involve os-net-config scripts under the condition that a new interface is added,, or an existing interface is deleted. Upon such operation, all the interfaces on the running host will be restarted even if it's not changed. The change we made was to make sure vlan interface won't be interrupted.

The design of os-net-config is idempotent. That means that if there are no differences between the configuration specified in the config template and the current system configuration then no interfaces are restarted. It sounds like your configuration is different in some way that is restarting all interfaces, so probably there is something in the ifcfg files which has been added by hand or when you are calling os-net-config you are using a different configuration than the one which was applied before.

It may help with troubleshooting to run os-net-config with the --debug flag, that will show you the changes that are different in the applied config and the current config. If you are making changes to an interface where the VLAN is located, or if the VLAN is on a bridge that is attached to an interface, then it is necessary to restart the VLAN in order to ensure proper functionality.

You can also use --debug with --noop to see what the config differences are without applying the interface or restarting anything.

Finally, you can use --no-activate to install the configuration without restarting any interfaces, and then manually restart only the interfaces you know need to be restarted.

If you would like more help with this please run os-net-config with --debug and --noop and attach the output to a comment. Thank you.

PR #14, opened 2024-03-07
Allow DNS/IP rule configuration only once

karthiksundaravel:
In case of rerun of os-net-config for NIC Partitioning use case, the dns and IP rule configurations could be applied 2 times (for sriov_pf and then for other objects). It could remove the existing configurations during first apply, resulting in a temporary loss of these settings, hence to avoid the same, its configured only during the second parse.


os_net_config/tests/test_cli.py
```diff
@@ -210,6 +210,12 @@ def apply(self, cleanup=False, activate=True):
                                            % interface_yaml, exitcodes=(0,))
 
     def test_sriov_noop_output(self):
+        def test_update_sriov_pf_map(name, numvfs, noop, promisc=None,
```
dsneddon, line 213: I see that the method "test_update_sriov_pf_map" is defined in many places and then stubbed out.

I think we could move this to L49 and then only define it once and stub it out once. Do you agree?

os_net_config/impl_nmstate.py
```diff
@@ -1570,7 +1570,7 @@ def add_ib_child_interface(self, ib_child_interface):
         data[InfiniBand.CONFIG_SUBTREE] = config
         self.interface_data[ib_child_interface.name] = data
 
-    def apply(self, cleanup=False, activate=True):
+    def apply(self, cleanup=False, activate=True, config_rules_dns=True):
```
dsneddon, line 1573: We need to add a comment at L1583 that defines the config_rules_dns:

        :param config_rules_dns: A boolean that indicates if the rules should
            be applied. This makes sure that the rules are configured only if
            config_rules_dns is set to True.

dsneddon (approved):
I would really like to see a follow-up patch that addresses my comments, but I know this patch is needed ASAP and I don't see any reason to block it right now.

PR #16, opened 2024-03-14
Reduce the definition of test stubs for update_sriov_pf_map

karthiksundaravel:
The stub for update_sriov_pf_map is defined for many test cases in test_objects.py. Reduced the redefinition of the same


dsneddon:
Thanks, this looks good, I appreciate the follow up.

vcandapp (approved):
lgtm

PR #17, opened 2024-04-02
Mismatch in table-id, metric shall be ignored for routes

karthiksundaravel:
With nmstate as provider, the existing routes are removed when the table-id or metric is not provided by the user. Added checks to skip removing the routes and re-adding them.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -506,9 +506,18 @@ class NmstateNetConfig(os_net_config.NetConfig):
 
         for c_route in curr_routes:
             no_metric = copy.deepcopy(c_route)
-            if NMRoute.METRIC in no_metric:
+            no_tableid = copy.deepcopy(c_route)
+            bare_min_route = copy.deepcopy(c_route)
+            if NMRoute.METRIC in bare_min_route:
+                del bare_min_route[NMRoute.METRIC]
                 del no_metric[NMRoute.METRIC]
-            if c_route not in reqd_route and no_metric not in reqd_route:
+            if NMRoute.TABLE_ID in bare_min_route:
+                del bare_min_route[NMRoute.TABLE_ID]
+                del no_tableid[NMRoute.TABLE_ID]
+            if c_route not in reqd_route and \
+                no_metric not in reqd_route and \
+                no_tableid not in reqd_route and \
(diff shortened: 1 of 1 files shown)


os_net_config/impl_nmstate.py
```diff
@@ -506,9 +506,18 @@ def generate_routes(self, interface_name):
...
+                del bare_min_route[NMRoute.METRIC]
                 del no_metric[NMRoute.METRIC]
-            if c_route not in reqd_route and no_metric not in reqd_route:
+            if NMRoute.TABLE_ID in bare_min_route:
+                del bare_min_route[NMRoute.TABLE_ID]
+                del no_tableid[NMRoute.TABLE_ID]
+            if c_route not in reqd_route and \
+                no_metric not in reqd_route and \
+                no_tableid not in reqd_route and \
+                bare_min_route not in reqd_route:
```
dsneddon, line 520: You are checking to see if the c_route exists in reqd_route with and without metric, with and without table_id, and without either. If no version of the route exists, then you append the route to routes. I think another way of doing this (possibly simpler?) would be to loop through reqd_route and strip the routes down to bare min, like this at 508:

bare_min_route = copy.deepcopy(c_route)
del bare_min_route[NMRoute.METRIC]
del bare_min_route[NMRoute.TABLE_ID]
bare_reqd_routes = []
for route in reqd_route:
____del route[NMRoute.METRIC]
____del route[NMRoute.TABLE_ID]
____bare_reqd_routes.append(route)
if bare_min_route not in bare_reqd_routes:
____c_route[NMRoute.STATE] = NMRoute.STATE_ABSENT
____routes.append(c_route)
____logger.info(f'Removing route {c_route}')
routes.extend(reqd_route)
return routes

However there may be a reason why we want to compare all variaties (with/without metric and/or table_id), but I'm not sure. Maybe the way you do it is fine in either case, but I'd like to understand why you chose to compare all versions of the route so please add a comment before I merge this pull request. Thanks!

dsneddon (commented):
Please take a look at my comment. I am wondering if there is an easier way to do this by simply stripping both the current route and the reqd routes down to bare minimum and comparing. Maybe it is neccessary to compare all versions, but I'm not seeing why. Thanks.

PR #19, opened 2024-04-10
Support IP Rules with route_table names

karthiksundaravel:
The below rules having route table names ``alternate`` can't be given to nmstate templates as a string. Instead it needs to be converted to a integer based on the map available at ``/etc/iproute2/rt_tables``.
- rule: "iif eno2 table alternate"
- rule: "from 10.0.2.0/24 table alternate"

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -797,7 +797,6 @@ class NmstateNetConfig(os_net_config.NetConfig):
                 value = get_route_options(route.route_options, 'table')
                 if value:
                     route.route_table = value
-
             if route.metric:
                 route_data[NMRoute.METRIC] = route.metric
             if route.ip_netmask:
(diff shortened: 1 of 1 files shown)


dsneddon:
Karthik, please give me your feedback on whether we should open a bug against nmstate to support table names. While I would be fine with merging this patch, the fact that nmstate does not accept table names and only accepts tables identified by integer seems like a bug. With the rules-<iface> files that were parsed by network-init-scripts it was possible to use either table IDs or names. Using names allows for consistency across systems where a specific ID is already in use. For example, suppose I have the table "DMZ3" with ID 33, and I want to apply similar routing rules to multiple systems, but one or more systems is already using table ID 33. I could write rules that refer to "DMZ3" and use the same rules on all systems regardless of ID. So this seems like a bug or a regression in nmstate to me.

karthiksundaravel:
> Karthik, please give me your feedback on whether we should open a bug against nmstate to support table names. While I would be fine with merging this patch, the fact that nmstate does not accept table names and only accepts tables identified by integer seems like a bug. With the rules- files that were parsed by network-init-scripts it was possible to use either table IDs or names. Using names allows for consistency across systems where a specific ID is already in use. For example, suppose I have the table "DMZ3" with ID 33, and I want to apply similar routing rules to multiple systems, but one or more systems is already using table ID 33. I could write rules that refer to "DMZ3" and use the same rules on all systems regardless of ID. So this seems like a bug or a regression in nmstate to me.

Yes, agreed. I have raised a JIRA card in nmstate - https://issues.redhat.com/browse/RHEL-32893. But until it is addressed in nmstate, os-net-config shall do the translation.

dsneddon:
> Yes, agreed. I have raised a JIRA card in nmstate - https://issues.redhat.com/browse/RHEL-32893. But until it is addressed in nmstate, os-net-config shall do the translation.

Thank you Karthik, I commented on the Jira card you linked to, let's track that and in the meantime I will happily merge this request. Let's add this to the growing list of features on the feature gap list between nmstate and os-net-config and perhaps we will be able to utilize os-net-config as a preprocessor or companion library alongside nmstate until features like this are supported natively.

PR #22, opened 2024-06-07
Update offload flag for legacy mode in SR-IOV PF

vcandapp:


os_net_config/impl_nmstate.py
```diff
@@ -1501,9 +1501,15 @@ def add_sriov_pf(self, sriov_pf):
         if sriov_pf.promisc:
             data[Interface.ACCEPT_ALL_MAC_ADDRESSES] = True
         if sriov_pf.link_mode == 'switchdev':
+            logger.info(f'enabling switchdev for sriov pf: {sriov_pf.name}')
```
karthiksundaravel, line 1504: Instead lets have the default value and an override when switchdev is enabled ?
data[Ethtool.CONFIG_SUBTREE] = {}
            data[Ethtool.CONFIG_SUBTREE][Ethtool.Feature.CONFIG_SUBTREE] = {
                'hw-tc-offload': False}

os_net_config/sriov_config.py
```diff
@@ -574,7 +574,7 @@ def trigger_udev_rules():
         raise
 
 
-def configure_switchdev(pf_name):
+def configure_switchdev(pf_name, link_mode='legacy'):
```
karthiksundaravel, line 577: I think, lets have another function to set the ethtool hw-tc-offload and invoke it from here. The same function can be called for legacy again.

karthiksundaravel (commented):
I have requested for couple of changes.

PR #32, opened 2024-07-19
Update the VF args for OvS / OvS user bridges

karthiksundaravel:
The default values of the VF parameters like Trust, Spoofcheck, Promisc are updated for OvS and OvS user bridges, Linux bonds. This update is not effectively applied on the nmstate provider's sriov_vf_data, leading to inconsistency between the sriov map and the VF configurations applied. Now the update of the default values are applied on sriov_vf_data.

### DIFF
--- a/os_net_config/tests/test_impl_nmstate.py
+++ b/os_net_config/tests/test_impl_nmstate.py
@@ -1394,9 +1394,10 @@ class TestNmstateNetConfig(base.TestCase):
         """
 
         ovs_obj = objects.object_from_json(yaml.safe_load(ovs_config))
+        bond_vf = ovs_obj.members[0]
+        for vf in bond_vf.members:
+            self.provider.add_sriov_vf(vf)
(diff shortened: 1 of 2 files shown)


karthiksundaravel:
recheck

os_net_config/tests/test_impl_nmstate.py
```diff
@@ -1394,9 +1394,9 @@ def test_get_vf_devname(device, vfid):
         """
 
         ovs_obj = objects.object_from_json(yaml.safe_load(ovs_config))
-        self.provider.add_bridge(ovs_obj)
         self.provider.add_sriov_vf(ovs_obj.members[0].members[0])
```
dsneddon, line 1397: Should we make a loop to include all members instead of just [0,1]?

dsneddon (commented):
Left a question for Karthik

raukadah:
recheck ci fixed!

vcandapp (approved):
lgtm

PR #35, opened 2024-07-25
sriov_config: fix race condition on boot

rjarry:
On boot, sriov_config.service runs in parallel with the probing of PCI devices. There is no synchronization point that we can use to run the service *after* all PCI devices have been probed. This leads to the service failing depending on the time it takes for devices to be probed:

```
12:31:30 os-net-config-sriov[1772]: FileNotFoundError: [Errno 2] No such file or directory: '/sys/class/net/ens2f0np0/device/sriov_numvfs'
...
12:31:43 kernel: mlx5_core 0000:17:00.0 ens2f0np0: renamed from eth0
```

The issue is that there is no UDEV rule to back this up when using NIC partitioning.
    
Add an explicit call to udevadm wait before trying to access the files. If the devices have already been probed, it will not cause any delay. On the other hand, if the devices haven't been probed yet or if the configuration is incorrect, the command will timeout after 120 seconds and raise an error.
    
Closes: https://issues.redhat.com/browse/OSPRH-8922

### DIFF
--- a/os_net_config/sriov_config.py
+++ b/os_net_config/sriov_config.py
@@ -170,6 +170,9 @@ def get_numvfs(ifname):
     :returns: int -- the number of current VFs on ifname
     :raises: SRIOVNumvfsException
     """
+    cmd = ["/usr/bin/udevadm", "wait", common.get_dev_path(ifname)]
+    logger.debug(f"{ifname}: Running command: {cmd}")
+    processutils.execute(*cmd)
     sriov_numvfs_path = common.get_dev_path(ifname, "sriov_numvfs")
     logger.debug(f"{ifname}: Getting numvfs for interface")
     try:
--- a/os_net_config/tests/test_sriov_config.py
+++ b/os_net_config/tests/test_sriov_config.py
@@ -33,8 +33,12 @@ class TestSriovConfig(base.TestCase):
         super(TestSriovConfig, self).setUp()
         rand = str(int(random.random() * 100000))
 
+        def execute_noop(*args, **kw):
+            pass
+
         tmpdir = tempfile.mkdtemp()
         self.stub_out('os_net_config.common.SYS_CLASS_NET', tmpdir)
+        self.stub_out('oslo_concurrency.processutils.execute', execute_noop)
 
         common._LOG_FILE = '/tmp/' + rand + 'os_net_config.log'
         sriov_config._UDEV_RULE_FILE = '/tmp/' + rand + 'etc_udev_rules.d'\


os_net_config/utils.py
```diff
@@ -37,6 +37,7 @@
 
 [Service]
 Type=oneshot
+ExecStartPre=/usr/bin/udevadm settle
```
dsneddon, line 40: There are still some possible edge cases where this would not work. The default timeout for "udevadm settle" is 120 seconds. We may want to increase this to something longer using --timeout=<seconds>, but we don't want to wait too long if the SR-IOV probing is done and udevadm is waiting for some other user rule to complete. If we know the sequence number of the rule we could use --seq-start=seqnum and --seq-stop=seqnum.

If we ran "udevadm settle" from within the os-net-config-sriov service could we do something like "udevadm settle --timeout=<seconds> --exit-if-exists=<file>" where the file is /sys/class/net/<ifname>/device/sriov_numvfs inside of get_numvfs in sriov_config.py in os-net-config when we call get_numvfs(pf_device['name']).

dsneddon (commented):
I added a comment suggesting we specify a long timeout for "udevadm settle" or perhaps even better we run "udevadm settle" from sriov_config.py with --exit-if-exists=<file> for the sriov_numvs file for each PF rather than delaying the start of the os-net-config-sriov service.

os_net_config/utils.py
rjarry: I've pushed an update. Tested it briefly and it seems to do the trick.

os_net_config/sriov_config.py
```diff
@@ -170,6 +170,9 @@ def get_numvfs(ifname):
     :returns: int -- the number of current VFs on ifname
     :raises: SRIOVNumvfsException
     """
+    cmd = ["/usr/bin/udevadm", "wait", common.get_dev_path(ifname)]
+    logger.debug(f"{ifname}: Running command: {cmd}")
+    processutils.execute(*cmd)
```
karthiksundaravel, line 175: Please add the processutils.execute under a try except block

karthiksundaravel (commented):
Thanks for the patch.

os_net_config/sriov_config.py
karthiksundaravel: We can add more readable error messages

os_net_config/sriov_config.py
```diff
@@ -171,6 +171,12 @@ def get_numvfs(ifname):
     :raises: SRIOVNumvfsException
     """
     sriov_numvfs_path = common.get_dev_path(ifname, "sriov_numvfs")
+    cmd = [
+        "/usr/bin/udevadm", "settle", "--timeout=120",
```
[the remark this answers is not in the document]
rjarry: This code is outdated. I have changed it to `udevadm wait`.

os_net_config/sriov_config.py
rjarry: It will be plenty readable:

```
Timed out for waiting devices being initialized.
```

It depends whether you want to crash or continue processing other vfs.

os_net_config/sriov_config.py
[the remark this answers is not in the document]
rjarry: I don't know about using a custom timeout. It seems like the default should be perfectly fine.

karthiksundaravel (approved):
LGTM. Thanks.

PR #37, opened 2024-07-31
Revert "Update the VF args for OvS / OvS user bridges"

dsneddon:
Reverts os-net-config/os-net-config#32


dsneddon:
There were issues with the master branch after this pull request was merged. Reverting this change.

PR #42, opened 2024-08-05
Exception logging shall also be stored in default log file

karthiksundaravel:
Allow `os-net-config-sriov` and `os-net-config-dcb` logs to be stored in the default log file. Also the exception logs shall be stored in the default log file.

Depends-On: https://github.com/os-net-config/os-net-config/pull/46

### DIFF
--- a/os_net_config/__init__.py
+++ b/os_net_config/__init__.py
@@ -315,7 +315,11 @@ class NetConfig(object):
         """
         logger.info('%s%s' % (self.log_prefix, msg))
         if not self.noop:
-            processutils.execute(cmd, *args, **kwargs)
+            out, err = processutils.execute(cmd, *args, **kwargs)
+            if err:
+                logger.error(f"stderr : {err}")
+            if out:
+                logger.debug(f"stdout : {out}")
 
     def write_config(self, filename, data, msg=None):
         msg = msg or "Writing config %s" % filename
--- a/os_net_config/sriov_config.py
+++ b/os_net_config/sriov_config.py
@@ -22,6 +22,7 @@
 # An entry point os-net-config-sriov is added for invocation of this module.
 
 import argparse
+import logging
 import os
 import pyudev
 import queue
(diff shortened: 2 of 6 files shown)


os_net_config/sriov_config.py
```diff
@@ -821,18 +822,17 @@ def parse_opts(argv):
     return opts
 
 
-def main(argv=sys.argv, main_logger=None):
+def main(argv=sys.argv):
     opts = parse_opts(argv)
-    if not main_logger:
-        main_logger = common.configure_logger(log_file=True)
-    common.logger_level(main_logger, opts.verbose, opts.debug)
+    logger = common.configure_logger(log_file=True)
```
vcandapp, line 827: Will this overwrite the module level logging (line 38) from sriov_config -> os-net-config ?

vcandapp (changes requested):

os_net_config/sriov_config.py
karthiksundaravel: Yes it will

os_net_config/__init__.py
```diff
@@ -315,7 +315,11 @@ def execute(self, msg, cmd, *args, **kwargs):
         """
         logger.info('%s%s' % (self.log_prefix, msg))
         if not self.noop:
-            processutils.execute(cmd, *args, **kwargs)
+            out, err = processutils.execute(cmd, *args, **kwargs)
```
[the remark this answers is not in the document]
karthiksundaravel: The uncaught exceptions will be handled by log_exceptions() in common.py

vcandapp (approved):
lgtm

karthiksundaravel:
recheck

dsneddon (changes requested):
We really need to modify this so that it doesn't quit with an exception if os-net-config is run by a normal user. There are valid reasons why a user might run os-net-config without root privileges, such as when running the -i interfaces command which just outputs live interfaces, or when running with --noop. Tox is failing because it can't open/create the os-net-config log file (which requires root in the chosen location). So if we modify the code to catch the permissions error and continue I think the tox tests may pass.

dsneddon (approved):
Tests are passing, we can follow up to allow non-root execution.

PR #44, opened 2024-08-09
Update schema for dpdkbond members

vcandapp:
Since there is a limitation from OvS for minimum 2 members required for dpdkbond members, updating
the schema to allow only 2 ports.
Set default failure for schema errors.
Added negative test cases for the same.

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -73,7 +73,7 @@ def parse_opts(argv):
         action='store_true',
         help="Exit with an error if configuration file validation fails. "
              "Without this option, just log a warning and continue.",
-        default=False)
+        default=True)
 
     parser.add_argument(
         '-d', '--debug',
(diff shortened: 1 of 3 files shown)


karthiksundaravel (approved):
lgtm. Thanks

dsneddon (commented):
I am not sure that an OVS DPDK bond can have only 2 members. It appears to me that the OVS source refers to RTE_MAX_ETHPORTS which is configured in the DPDK source as 'max_ethports' and defaults to 32 when compiling. So I think we should only have minitems: 2 with no maxitems and remove the test which checks to make sure there are no more than 2.

vcandapp:
> I am not sure that an OVS DPDK bond can have only 2 members. It appears to me that the OVS source refers to RTE_MAX_ETHPORTS which is configured in the DPDK source as 'max_ethports' and defaults to 32 when compiling. So I think we should only have minitems: 2 with no maxitems and remove the test which checks to make sure there are no more than 2.

Updated shcema for checking only min items

PR #45, opened 2024-08-19
Adding metric option to route

karthiksundaravel:


os_net_config/schema.yaml
```diff
@@ -180,6 +180,8 @@ definitions:
                 anyOf:
                     - $ref: "#/definitions/string_or_param"
                     - $ref: "#/definitions/int_or_param"
+            metric:
```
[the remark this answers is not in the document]
karthiksundaravel: Yes. I'll make the change

os_net_config/schema.yaml
```diff
@@ -196,6 +198,8 @@ definitions:
...
+                $ref: "#/definitions/int_or_param"
```
vcandapp, line 202: Both methods of specifying metrics is allowed now ? Currently we can do like this: https://github.com/os-net-config/os-net-config/blob/a1ec327a0654e435d5518e0b60b21d23434b45e6/etc/os-net-config/samples/routes.yaml#L17-L27

karthiksundaravel: That will lead to confusions if its specified at both places. Better to define in only one place.

PR #46, opened 2024-08-20
DCB config schema update

karthiksundaravel:
Modified the day2 configuration templates for dcb config where the ``name`` is replaced with ``device``


abhiramnarayana:
Integrated and verified changes on DCB setup for Day 1, Day2 configurations and reboot scenarios. Its working fine.

dsneddon (approved):
I didn't test, but I went through each file and verified that each instance of 'name' seems to have been changed to 'device' in the proper places, and the sample was updated. LGTM and Karthik I can help with documentation but it needs to be done. Thanks so much!

vcandapp (approved):
lgmt

PR #47, opened 2024-08-22
Add ovs_internal to prevent OVS-DPDK kernel iface

dsneddon:
This patch adds the ovs_interface configuration option for OVS User Bridges. It defaults to True, but if set to False the LINKSTATUS=down statement is added to the ifcfg file for the OvsUserBridge. This statement will prevent the internal OVS interface from coming up, which prevents OVS-DPDK from flooding the kernel with BUM traffic, which can hurt performance.


karthiksundaravel (approved):
lgtm. thanks

vcandapp (approved):
lgtm

PR #54, opened 2024-09-03
Adoption from ifcfg to nmstate provider

karthiksundaravel:
Add support for reconfiguring the network interfaces from ifcfg provider to nmstate in existing deployments. The steps are
1. Run ifdown on all intrerfaces
2. Move ifcfg files from /etc/sysconfig/network-scripts/ to /var/lib/os-net-config/ifcfg-purge/
3. Run ```nmcli con reload```
4. Mark the device as absent in nmstate provider
5. Configure the device with nmstate provider
6. Wipe the residue files from ifcfg provider


dsneddon (approved):
Very nice. There seems to be a lot of attention to detail in this patch and I'm very happy with it. We need to get it properly rebased then we can merge.

os_net_config/__init__.py
```diff
@@ -128,34 +188,69 @@ def add_route_table(self, route_table):
...
         """
         raise NotImplementedError("add_ovs_user_bridge is not implemented.")
 
+    def del_ovs_user_bridge(self, bridge):
+        """Delete an OvsUserBridge object to the net config object.
+
+        :param bridge: The OvsUserBridge object to be deleted.
+        """
+        raise NotImplementedError("del_ovs_user_bridge is not implemented.")
+
```
vcandapp, line 253: Is there no valid use case for del_linux_bridge () ? Controller nodes might have created linux_bridges

vcandapp (changes requested):

PR #57, opened 2024-09-04
Roll back to previous provider incase of migration failure

karthiksundaravel:
In cases where migration of provider fails, the network configurations shall be rolled back to the previous provider.

Depends-On: https://github.com/os-net-config/os-net-config/pull/54


raukadah:
recheck

dsneddon:
recheck

dsneddon (approved):
LGTM

PR #70, opened 2024-09-13
Raise exception when switchdev/vDPA is enabled with nmstate provider

karthiksundaravel:
Switchdev and vDPA are not supported yet with nmstate provider. ConfigurationError exception is raised when these features are used with nmstate provider.


dsneddon (approved):
LGTM

PR #80, opened 2024-09-16
Ci fix

vcandapp:
Cherry picking CI fix, as the tox error was encountered in downstream trunk-patches


karthiksundaravel:
PR raised incorrectly. Abandoning it

PR #82, opened 2024-09-16
DNM - Allow os-net-config to run on RHEL8

karthiksundaravel:
RHEL8 includes older version of libnmstate which does not include certain modules or class. This resulted in import error for the same. The import errors deny os-net-config the possibility to run with ifcfg provider.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -14,32 +14,34 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
-from libnmstate import error
-from libnmstate import gen_diff
-from libnmstate import netapplier
-from libnmstate import netinfo
-from libnmstate.schema import Bond
-from libnmstate.schema import BondMode
-from libnmstate.schema import DNS
-from libnmstate.schema import Ethernet
-from libnmstate.schema import Ethtool
-from libnmstate.schema import InfiniBand
-from libnmstate.schema import Interface
-from libnmstate.schema import InterfaceIPv4
-from libnmstate.schema import InterfaceIPv6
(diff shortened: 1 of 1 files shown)


os_net_config/impl_nmstate.py
```diff
@@ -14,32 +14,34 @@
...
+    from libnmstate.schema import OVSInterface
+    from libnmstate.schema import Route as NMRoute
+    from libnmstate.schema import RouteRule as NMRouteRule
+    from libnmstate.schema import VLAN
+    import logging
+    import netaddr
+    import re
+    import yaml
+except:
+    pass
```
nate-johnston, line 44: I think there should be a log message here indicating that the libnmstate version is old and therefore the nmstate provider will not work on this version.

os_net_config/impl_nmstate.py
```diff
@@ -14,32 +14,34 @@
...
+    from libnmstate.schema import OVSBridge
+    from libnmstate.schema import OvsDB
+    from libnmstate.schema import OVSInterface
+    from libnmstate.schema import Route as NMRoute
+    from libnmstate.schema import RouteRule as NMRouteRule
+    from libnmstate.schema import VLAN
+    import logging
+    import netaddr
+    import re
+    import yaml
```
nate-johnston, line 42: Can we pare down this list to the problematic set - or perhaps even just the libnmstate.schema items?  I doubt we are going to have an import issue with 're' or 'yaml'.

nate-johnston (changes requested):
I know this is DNM but I have a few improvements to suggest

karthiksundaravel:
Abandoned

PR #83, opened 2024-09-16
Allow os-net-config to run on RHEL8

rjarry:


nate-johnston:
The import originally causing issues was `from libnmstate.schema import Ethtool` but I don't see that affected in this change, does that need additional handling or does this obviate the import error on RHEL 8 in some way?

dsneddon (approved):
I like these changes

karthiksundaravel:
Thanks Robin, we need to make the tox jobs work .

os_net_config/cli.py
```diff
@@ -202,17 +199,24 @@ def main(argv=sys.argv, main_logger=None):
                               "system.")
             return 1
 
+    try:
```
karthiksundaravel, line 201: Please move this block after line #312. Provider needs to be initialised after purge provider, so that the migration sequence is taken care.

karthiksundaravel (commented):
Getting the error
root@dell-r450-nfv2023-03:~/rjarry/os-net-config lazy# os-net-config -d -p ifcfg -c ~/templates/iface_1.yaml 
2024-09-17 08:19:03.920 INFO os_net_config.main Using config file at: /root/templates/iface_1.yaml
2024-09-17 08:19:03.920 ERROR os_net_config.main cannot load provider ifcfg: ModuleNotFoundError("No module named 'impl_ifcfg'")

os_net_config/cli.py
```diff
@@ -165,6 +166,11 @@ def get_sriovpf_member_of_bond_ovs_port(obj):
     return net_devs_list
 
 
+def load_provider(name: str, noop: bool, root_dir: str) -> NetConfig:
+    mod = importlib.import_module(f'impl_{name}', package='.')
```
karthiksundaravel, line 169: mod = importlib.import_module(f'os_net_config.impl_{name}', package='.')

os_net_config/cli.py
```diff
@@ -16,17 +16,15 @@
...
 import argparse
+import importlib
 import json
 import os
 import sys
 import yaml
 
+from os_net_config import NetConfig
 from os_net_config import common
 from os_net_config import dcb_config
```
karthiksundaravel: This needs to be done after importing common to avoid the pep8 error.

os_net_config/cli.py
rjarry: I have fixed it.

os_net_config/cli.py
rjarry: Done

rjarry:
> The import originally causing issues was `from libnmstate.schema import Ethtool` but I don't see that affected in this change, does that need additional handling or does this obviate the import error on RHEL 8 in some way?

Yes, since the import of the whole impl_nmstate.py module is now only done if the provider is nmstate, none of the libnmstate symbols are imported unless required.

os_net_config/cli.py
rjarry: done

karthiksundaravel (approved):
LGTM. Thanks.

karthiksundaravel:
Its verified locally on the OSP17 nodes for the impacted use cases. It clearly passes the regression on RHEL9.

PR #84, opened 2024-09-17
providers: lazily load NetConfig implementations

karthiksundaravel:
On some environment, libnmstate is too old and we cannot upgrade it. However, it shouldn't prevent from running os-net-config with the ifcfg provider.

Only import impl_nmstate if the requested provider actually is nmstate. If there is a missing module, an explicit error will be raised.

(cherry picked from commit 6dffeea98d0999ab509043f1bec524ad5bc9aeac)

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -16,6 +16,7 @@
 
 
 import argparse
+import importlib
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


vcandapp (approved):
lgtm

PR #86, opened 2024-09-19
Set the default provider as nmstate

karthiksundaravel:
When the environment has the libnmstate support available with a minimum desired version, the default provider shall be set to nmstate

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -176,6 +176,23 @@ def load_provider(name, noop, root_dir):
     return provider_class(noop=noop, root_dir=root_dir)
 
 
+def is_nmstate_available():
+    try:
+        import libnmstate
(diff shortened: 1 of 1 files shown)


nate-johnston:
Can we make sure this gets in the version for 18.0 but not the version shipped with 17.1.4?  Open to other opinions but just my opinion I don't think we're ready for this in 17.1.4.

karthiksundaravel:
Yes agreed, that's the plan.

dsneddon (approved):
This is a better test of the presence and compatible version of nmstate. 

However at line 253 where we check for "if opts.provider is not None:" we should really move that test before the provider checks, that way we can just parse the interface list and return before we need to do anything which requires super user. It should be moved before the provider checks. We can do that in a follow-up and it's low priority.

PR #88, opened 2024-09-21
Handle ovs-extra for ovs dpdk ports and bonds

karthiksundaravel:
Handle the below ovs_extra settings for dpdk ports/bonds
  - set port dpdkbond0 bond_mode=balance-slb
  - set port dpdkbond1 bond_updelay=1000
  - set port dpdkbond1 other_config:bond-detect-mode=miimon
  - set port dpdkbond1 other_config:bond-miimon-interval=100
  - set Interface dpdk0 options:n_rxq_desc=2048
  - set Interface dpdk0 options:n_txq_desc=2048

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -130,6 +130,9 @@ def is_dict_subset(superset, subset):
 
 
 def _add_sub_tree(data, subtree):
+    if data is None:
+        msg = 'Subtree can\'t be created on None Types'
+        raise os_net_config.ConfigurationError(msg)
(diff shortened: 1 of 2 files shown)


dsneddon (approved):
This looks fairly complete, and it should add the needed extra params for proper operation of OVS in this context.

dsneddon:
I want to make sure we are recording this in the right place to get downstream approval. Is this the right BZ for 17?
https://bugzilla.redhat.com/show_bug.cgi?id=2314230
What is the Jira issue related to this patch?

PR #89, opened 2024-09-22
Report failure when provider fails to configure network

karthiksundaravel:
When the provider fails to configure the network, it reports an appropriate exception. The exception is handled so that the migration cases like rollback is supported, but the exception needs to be raised for non migration cases.


vcandapp (approved):
lgtm

dsneddon:
Is this the BZ for 17?

https://bugzilla.redhat.com/show_bug.cgi?id=2313764

What is the Jira for this patch?

PR #90, opened 2024-09-22
Move stubs that are repeated several times

karthiksundaravel:
Some of the stubs are repeated for some of the tests in test_impl_nmstate.py. These stubs are moved to the setUp() of the respective class, so that the stub implementation is not repetitive.


dsneddon (approved):
This is much cleaner, thanks

PR #91, opened 2024-09-24
Report failure when provider fails to configure network

karthiksundaravel:
When the provider fails to configure the network, it reports an appropriate exception. The exception is handled so that the migration cases like rollback is supported, but the exception needs to be raised for non migration cases.

(cherry picked from commit 886b2a5578036f15b98a5b97a7e1ba53298cf555)


dsneddon (approved):
LGTM

PR #94, opened 2024-09-24
DCB configurations over OVS DPDK BOND

abhiramnarayana:
For the DCB configurations over a dpdkbond, currently the configs needs to be repeated under all the dpdk bond members (interfaces). This patch adds support in os-net-config to be able to apply DCB configurations directly under the object of type ovs dpdk bond.


karthiksundaravel:
recheck

dsneddon (approved):
LGTM

karthiksundaravel:
test cases needs to be added, but that could be a follow up patch. Thanks

PR #96, opened 2024-10-01
Change DNS without restarting interface

dsneddon:
With this change the DNS servers and search domains can be changed in place without restarting the interface when using ifcfg provider. This change also does some code cleanup for readability and updates the string handling to be more Python3 compliant in some places.

### DIFF
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -289,7 +289,9 @@ class IfcfgNetConfig(os_net_config.NetConfig):
         """Return list of commands needed to implement changes.
 
            Given ifcfg data for an interface, return commands required to
-           apply the configuration using 'ip' commands.
+           apply the configuration using 'ip' commands. Changes to DNS
+           servers or DNS search domains are made directly in resolv.conf
+           using 'sed' and/or 'echo' directly on /etc/resolv.conf.
 
         :param dev_name: The name of the int, bridge, or bond
         :type dev_name: string
@@ -340,13 +342,13 @@ class IfcfgNetConfig(os_net_config.NetConfig):
             if changes['DOMAIN'] == 'added':
                 commands.append(f"echo 'search {new}' >> /etc/resolv.conf")
             elif changes['DOMAIN'] == 'modified':
-                commands.append(f"sed -i -e 's/{old}/{new}/g' "
-                                "/etc/resolv.conf")
+                commands.append(f"sed -i -e 's/search[ ]*{old}/search {new}/g'"
+                                " /etc/resolv.conf")
             elif changes['DOMAIN'] == 'removed':
-                commands.insert("sed -i '/search/d' /etc/resolv.conf")
-        if 'DNS1' in changes:
-            # Remove nameservers from /etc/resolv.conf
-            commands.append("sed -i '/nameserver/d' /etc/resolv.conf")
+                commands.insert("sed -i '/^search/d' /etc/resolv.conf")
+        if 'DNS1' in changes or 'DNS2' in changes:
+            # Remove all nameservers from /etc/resolv.conf
+            commands.append("sed -i '/^nameserver/d' /etc/resolv.conf")
             # Add first new nameserver in /etc/resolv.conf
             if changes['DNS1'] == 'added' or changes['DNS1'] == 'modified':
                 ns1 = data_values['DNS1']
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -167,25 +167,25 @@ class IfcfgNetConfig(os_net_config.NetConfig):
            Return the keys and values without quotes.
            """
         ifcfg_values = {}
-        for line in ifcfg_data.split("\n"):
-            if not line.startswith("#") and line.find("=") > 0:
-                k, v = line.split("=", 1)
-                ifcfg_values[k] = v.strip("\"'")
+        for line in ifcfg_data.split('\n'):
+            if not line.startswith('#') and line.find('=') > 0:
+                k, v = line.split('=', 1)
+                ifcfg_values[k] = v.strip('"\'')
         return ifcfg_values
 
     def parse_ifcfg_routes(self, ifcfg_data):
         """Break out the individual routes from an ifcfg route file."""
         routes = []
-        for line in ifcfg_data.split("\n"):
-            if not line.startswith("#"):
+        for line in ifcfg_data.split('\n'):
+            if not line.startswith('#'):
                 routes.append(line)
         return routes
 
     def parse_ifcfg_rules(self, ifcfg_data):
         """Break out the individual rules from an ifcfg rule file."""
         rules = []
-        for line in ifcfg_data.split("\n"):
-            if not line.startswith("#"):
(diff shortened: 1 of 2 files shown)


os_net_config/impl_ifcfg.py
```diff
@@ -309,56 +303,58 @@ def iproute2_apply_commands(self, device_name, filename, data):
...
+        if 'DOMAIN' in changes:
+            old = file_values['DOMAIN']
+            new = data_values['DOMAIN']
+            if changes['DOMAIN'] == 'added':
+                commands.append(f"echo 'search {new}' >> /etc/resolv.conf")
+            elif changes['DOMAIN'] == 'modified':
+                commands.append(f"sed -i -e 's/{old}/{new}/g' "
+                                "/etc/resolv.conf")
+            elif changes['DOMAIN'] == 'removed':
+                commands.insert("sed -i '/search/d' /etc/resolv.conf")
```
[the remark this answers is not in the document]
dsneddon: Yes, that's better, I'll make that change, thanks.

os_net_config/impl_ifcfg.py
```diff
@@ -309,56 +303,58 @@ def iproute2_apply_commands(self, device_name, filename, data):
...
+        if 'DNS1' in changes:
             # Remove nameservers from /etc/resolv.conf
-            changes.append("sed -i '/nameserver/d' /etc/resolv.conf")
+            commands.append("sed -i '/nameserver/d' /etc/resolv.conf")
```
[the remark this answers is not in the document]
dsneddon: There is an issue with the method you are using, and that's that the "sed -i '/^nameserver[]/d resolv.conf" will remove all nameservers. So if we go from two nameservers to one it could cause issues. Rather than that if either one changes I will remove both and then add one or both from the new data.

os_net_config/impl_ifcfg.py
```diff
@@ -340,13 +342,13 @@ def iproute2_apply_commands(self, dev_name, filename, data):
...
+                                " /etc/resolv.conf")
             elif changes['DOMAIN'] == 'removed':
-                commands.insert("sed -i '/search/d' /etc/resolv.conf")
-        if 'DNS1' in changes:
-            # Remove nameservers from /etc/resolv.conf
-            commands.append("sed -i '/nameserver/d' /etc/resolv.conf")
+                commands.insert("sed -i '/^search/d' /etc/resolv.conf")
+        if 'DNS1' in changes or 'DNS2' in changes:
+            # Remove all nameservers from /etc/resolv.conf
+            commands.append("sed -i '/^nameserver/d' /etc/resolv.conf")
```
karthiksundaravel, line 351: Can we use the same sed command pattern for DNS as well

os_net_config/impl_ifcfg.py
```diff
@@ -340,13 +342,13 @@ def iproute2_apply_commands(self, dev_name, filename, data):
...
-                commands.append(f"sed -i -e 's/{old}/{new}/g' "
-                                "/etc/resolv.conf")
+                commands.append(f"sed -i -e 's/search[ ]*{old}/search {new}/g'"
+                                " /etc/resolv.conf")
             elif changes['DOMAIN'] == 'removed':
-                commands.insert("sed -i '/search/d' /etc/resolv.conf")
-        if 'DNS1' in changes:
-            # Remove nameservers from /etc/resolv.conf
-            commands.append("sed -i '/nameserver/d' /etc/resolv.conf")
+                commands.insert("sed -i '/^search/d' /etc/resolv.conf")
```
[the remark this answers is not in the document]
dsneddon: This was modified. I need to check if either DNS1 or DNS2 is modified, and if so I need to remove both and add both new values. This was because I ran into issues if an IP moved from DNS1 to DNS2 or the reverse.

os_net_config/impl_ifcfg.py
dsneddon: I don't believe so, I found edge cases where this didn't work. For example, let's say we have the following before update:
DNS1=10.0.1.14
DNS2=10.0.2.15
And the following after update:
DNS1=10.0.2.15
DNS2=10.0.3.16
When I ran the sed change to modify DNS1 in place, it would then appear like this:
DNS1=10.0.2.15
DNS2=10.0.2.15
When I then ran the update in place for DNS2 it would update both DNS1 and DNS2. I ran into other edge cases as well, and the only certain way of modifying both DNS1 and DNS2 was to remove both and add them back.

dsneddon: I found some edge cases where this didn't work. For example if there were two DNS servers before update and one after. I suppose I could do this:
commands.append("sed -i '/^nameserver {old}/d' /etc/resolv.conf")

If DNS 2 was removed.

However I ran into some edge cases where there were issues. If DNS1 was moved to DNS2, and DNS1 was modified, I could run into a situation where /etc/resolv.conf looked like this:

ifcfg before change:
DNS1=10.0.1.14
DNS2=10.0.2.15
and after change:
DNS1=10.0.2.15
DNS2=10.0.3.16

Then after the nameserver for DNS1 was changed in place the /etc/resolv.conf would look like this:

nameserver 10.0.2.15
nameserver 10.0.2.15

So then if I ran "sed -i '^nameserver 10.0.2.15/d' /etc/resolv.conf" it would remove both nameservers and then add back only one, so DNS1 would not be added to the file and we would end up with only DNS2.

I found it was best to remove all nameservers and then immediately add the new ones. Notice that these commands are queued up immediately after one another:
sed -i '/^nameserver/d' /etc/resolv.conf
echo -e 'nameserver <ns1>' >> /etc/resolv.conf
echo -e 'nameserver <ns2> ' >> /etc/resolv.conf
So the chances of a nameserver lookup happening between the remove of both nameservers and the addition of the new nameservers is extremely small. Since this worked for all use cases I threw at it I didn't think it was worth it to risk an edge case hitting us if we were trying to modify in-place.

dsneddon:
Karthik, I think the changes in patchset 4 are the best we can do for the moment. The impact of removing both DNS servers and adding both new DNS servers is extremely small, but it avoids some edge cases that could prevent both DNS servers from being added properly. I think we should approve and get these patches cherry-picked downstream ASAP, we will run into serious issues if these changes are not merged downstream by Friday. Thanks!

os_net_config/impl_ifcfg.py
```diff
@@ -259,43 +259,42 @@ def ifcfg_requires_restart(self, filename, new_data):
...
         # Certain changes can be applied without restarting the interface
         permitted_changes = [
-            "IPADDR",
-            "NETMASK",
-            "MTU",
-            "ONBOOT",
-            "ETHTOOL_OPTS"
+            'IPADDR', 'NETMASK',
+            'MTU', 'ONBOOT', 'ETHTOOL_OPTS',
+            'DOMAIN', 'DNS1', 'DNS2'
```
vcandapp, line 271: If IPADDR is supported, shouldn't "IPV6ADDR" also be allowed without reboot ?

dsneddon: Maybe in a follow up

vcandapp (approved):
lgtm

PR #97, opened 2024-10-02
DNM - Perform DPDK binding after VF configuration

karthiksundaravel:
In case of NIC partitioning using nmstate provider the DPDK binding shall be performed after the configuration of VFs, so as to allow proper configuration of the VFs and the DPDK driver bindings.


karthiksundaravel:
Abandoning it and changed the approach. raised PR#119

PR #98, opened 2024-10-02
SR-IOV configurations are made one after other for nmstate provider

karthiksundaravel:
SR-IOV configurations are applied one after the other. Its found that applying the SR-IOV configurations of all the NICs in one shot using nmstate results in failure 'Unable to configure VFs, other operation is pending` for i40e drivers.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -101,7 +101,12 @@ def is_dict_subset(superset, subset):
     if superset and subset:
         for key, value in subset.items():
             if key not in superset:
-                return False
+                # Items which are empty or false
+                # shall be considered as absent
+                if value:
+                    return False
+                else:
+                    continue
             if isinstance(value, dict):
                 if not is_dict_subset(superset[key], value):
                     return False
@@ -118,7 +123,14 @@ def is_dict_subset(superset, subset):
                 except TypeError:
                     for item in value:
                         if item not in superset[key]:
-                            return False
+                            if isinstance(item, dict):
+                                for s_items in superset[key]:
+                                    if is_dict_subset(s_items, item):
+                                        break
+                                else:
+                                    return False
+                            else:
+                                return False
             elif isinstance(value, set):
                 if not value <= superset[key]:
                     return False
@@ -240,6 +252,8 @@ class NmstateNetConfig(os_net_config.NetConfig):
         self.sriov_pf_data = {}
         self.sriov_vf_data = {}
         self.migration_enabled = False
+        self.apply_pf = False
+        self.apply_vf = False
         self.initial_state = netinfo.show_running_config()
         self.__dump_key_config(self.initial_state,
                                msg="Initial network settings")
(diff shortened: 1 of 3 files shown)


dsneddon (approved):

dsneddon (commented):
LGTM

dsneddon (dismissed):

karthiksundaravel:
recheck

rjarry (changes requested):
This seems like you are working around an nmstate or Network Manager bug. They should not try to apply the configuration again if nothing has changed.

If the root cause bug cannot be fixed soon enough and this workaround is needed. Please state it loudly in comments and in the commit message, referencing the original ticket and say this patch needs to be reverted as soon as the bug is fixed.

Cheers.

karthiksundaravel:
> This seems like you are working around an nmstate or Network Manager bug. They should not try to apply the configuration again if nothing has changed.
> 
> If the root cause bug cannot be fixed soon enough and this workaround is needed. Please state it loudly in comments and in the commit message, referencing the original ticket and say this patch needs to be reverted as soon as the bug is fixed.
> 
> Cheers.

Updated the comments and commit message.

dsneddon (commented):
Just one question about if we have a tracking issue for the RFE(s) for nmstate

dsneddon (dismissed):
Thanks for the updates, this looks good now.

os_net_config/impl_nmstate.py
```diff
@@ -316,27 +330,73 @@ def get_vf_config(self, sriov_vf):
...
+
+        :param activate: A boolean which indicates if the config should
+            be activated by applying the desired state.
+        '''
+        pf_config = []
+
+        # FIXME: The desired state of the PF's are applied one after the
+        # other, so as to avoid driver errors. This needs to be addressed
+        # from nmstate instead. Also the comparison of the current state
+        # vs desired state needs to be performed in nmstate.
```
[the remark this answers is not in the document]
karthiksundaravel: Added the Jira here

rjarry (changes requested):
You are doing two things in one single commit.

1. Apply SRIOV configs one at a time
2. Only apply configs if something changed

Could you split that in two separate changes? Since 2. is a way to workaround a nmstate bug/quirk, we will be able to revert it eventually.

Thanks

os_net_config/impl_nmstate.py
```diff
@@ -248,6 +248,8 @@ def __init__(self, noop=False, root_dir=''):
         self.sriov_pf_data = {}
         self.sriov_vf_data = {}
         self.migration_enabled = False
+        self.apply_pf = False
+        self.apply_vf = False
```
rjarry, line 252: Could you add a comment describing what these two booleans are for?

The variable names are probably not great. How about:

```python
self.need_pf_config = False
self.need_vf_config = False
```

rjarry (changes requested):

os_net_config/impl_nmstate.py
karthiksundaravel: Added the comments and renamed the variable names

rjarry (approved):

PR #100, opened 2024-10-03
Do not fail on schema validation errors

karthiksundaravel:

- type: ovs_bridge
  name: br-ex
  use_dhcp: false
- type: ovs_bridge
  name: br-vlan
  use_dhcp: false

This commit restore the previous behavior, where we emit a warning. It is left to the application/user to pass the --exit-on-validation-errors option at runtime if required.

BZ issue: https://bugzilla.redhat.com/show_bug.cgi?id=2316090

(cherry picked from commit 62a9dba3c152f41440e250c5c33d55944514b2db)

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -86,7 +86,7 @@ def parse_opts(argv):
         action='store_true',
         help="Exit with an error if configuration file validation fails. "
              "Without this option, just log a warning and continue.",
-        default=True)
+        default=False)
 
     parser.add_argument(
         '-d', '--debug',


vcandapp (approved):

vcandapp:
@karthiksundaravel / @dsneddon , if schema validation fails - shouldn't the safe_defaults be triggered ? Is ignoring schema validation better or removing 'members' as mandatory in https://github.com/os-net-config/os-net-config/blob/master/os_net_config/schema.yaml#L628 better ?

vcandapp:
@karthiksundaravel / @dsneddon, for master - we can revert the PR https://github.com/os-net-config/os-net-config/pull/99

dsneddon:
I think we want to fail in validation errors, but we need a new PR which removes the requires for members

dsneddon:
That way we can create bridges without physical ports and not fail schema validation, so the DHCP all interfaces will not run.

dsneddon:
So maybe we don't allow validation errors, but instead we fix the schema so that it is possible to create a bridge without members and there will be no error that we have to skip past.

dsneddon:
So yes, please continue to trigger safe defaults if validation errors are encountered, but let's create a new patch that allows no members on the bridge.

karthiksundaravel:
Abandonned

PR #102, opened 2024-10-04
Allow empty members for ovs_bridge

vcandapp:
Instead of allowing schema error (PR#100), updating schema for ovs_bridge to allow empty members:

type: ovs_bridge
name: br-ex
use_dhcp: false
type: ovs_bridge
name: br-vlan
use_dhcp: false

Added test case for same

Fixes issue: https://bugzilla.redhat.com/show_bug.cgi?id=2316090


karthiksundaravel (commented):
lgtm

vcandapp:
@lmiccini , please review this

dsneddon (approved):

karthiksundaravel (approved):
lgtm

PR #107, opened 2024-10-07
ci: add script to check git commits

rjarry:
Recently, there were multiple occasions where the git commits were not following the expected standards. Add a github action that will check every commit in all pull requests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+---
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+
+jobs:
+  check-commits:
+    if: ${{ github.event.pull_request.base.sha && github.event.pull_request.head.sha }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # force fetch all history
+      - run: git config --global --add safe.directory $PWD
+      - run: ./check-commits "${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }}"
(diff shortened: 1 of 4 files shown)


dsneddon (commented):
I like where this is going, but apart from referencing this action script where can I find a formatting guide which specifies the conventions that should be followed? I don't think I've ever seen an OpenStack contributing guide that had rules quite as strict as those implemented here, so where did these conventions come from?

rjarry:
The actual conventions are written up here:

https://git.sr.ht/~rjarry/aerc/tree/master/item/CONTRIBUTING.md
https://github.com/DPDK/grout/blob/main/CONTRIBUTING.md

karthiksundaravel (approved):
LGTM Thanks.

.github/workflows/check-commits
```diff
@@ -0,0 +1,134 @@
+#!/bin/sh
+# SPDX-License-Identifier: Apache-2.0
+# Copyright (c) 2024 Robin Jarry
+
```
vcandapp, line 4: @karthiksundaravel / @dsneddon - can we use copyright statement like this ? We have these in other files - eg; -https://github.com/os-net-config/os-net-config/blob/master/os_net_config/sriov_bind_config.py#L1

vcandapp (changes requested):

.github/workflows/check-commits
dsneddon: Perhaps @rjarry would be willing to make this available under the same license as the rest of os-net-config?

rjarry: This is the same license. Only a different copyright. I removed it. It makes no sense to add a license to such kind of script :)

karthiksundaravel (approved):
Happy with the changes. Thanks.

vcandapp (approved):
LGTM, thanks

PR #108, opened 2024-10-07
Allow empty members for ovs_bridge

vcandapp:
Instead of allowing schema error (PR#100), updating schema for ovs_bridge to allow empty members:

type: ovs_bridge
name: br-ex
use_dhcp: false
type: ovs_bridge
name: br-vlan
use_dhcp: false

Added test case for same

(cherry picked from commit 3ce3e2cefd30ced972148a5151ddf89a779348e3)


karthiksundaravel (approved):
LGTM. Thanks

karthiksundaravel (dismissed):
lgtm

karthiksundaravel (approved):
lgtm

PR #110, opened 2024-10-07
ifcfg-dns: Allow DNS changes without restarting

dsneddon:
This change allows DNS servers and search domains to be modified on the fly without restarting interfaces. The DNS nameservers and search domains are written to the ifcfg file and /etc/resolv.conf will be updated in place.

This change also includes some code cleanup and updates string formatting in some places to match Python3 style.

Link: https://bugzilla.redhat.com/2315341
Fixes: https://issues.redhat.com/browse/OSP-32857
Fixes: https://issues.redhat.com/browse/OSP-32858

### DIFF
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -167,25 +167,25 @@ class IfcfgNetConfig(os_net_config.NetConfig):
            Return the keys and values without quotes.
            """
         ifcfg_values = {}
-        for line in ifcfg_data.split("\n"):
-            if not line.startswith("#") and line.find("=") > 0:
-                k, v = line.split("=", 1)
-                ifcfg_values[k] = v.strip("\"'")
+        for line in ifcfg_data.split('\n'):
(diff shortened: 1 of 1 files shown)


karthiksundaravel (approved):
Thanks. LGTM

karthiksundaravel:
If DNS is removed/added, then the new ifcfg has changes w.r.t PEERDNS which triggers the ifdown/ifup.
If DNS is not present in ifcfg and DOMAIN is removed now, then there is no update to the resolv.conf as well.
I understand that these are corner and specific cases that might not be required in the current context.

dsneddon:
> If DNS is removed/added, then the new ifcfg has changes w.r.t PEERDNS which triggers the ifdown/ifup. If DNS is not present in ifcfg and DOMAIN is removed now, then there is no update to the resolv.conf as well. I understand that these are corner and specific cases that might not be required in the current context.

It is not expected that changes to PEERDNS can be done without restart. The PEERDNS setting affects DHCP so the interface will be restarted to obtain a new DHCP response.

PR #112, opened 2024-10-09
 noop: Run os-net-config without applying the changes

karthiksundaravel:
When os-net-config is run with noop, its failing due to permission issues or trying to write files and configure devices. When `noop` is used, there shall be no changes made to the machine, while the configuration files or templates derived shall be output on the console. This is done so that the template/compatibility issues of the config.yaml could be validated on a given hardware without changing the system files or configuration.


dsneddon (approved):
Very good change

dsneddon (approved):
Very good change

dsneddon (approved):

PR #113, opened 2024-10-14
Add bond option fail_over_mac

karthiksundaravel:
add into linux bond options fail_over_mac


SharonBX:
linux_bond: Add fail_over_mac support to the linux bonding options.

karthiksundaravel (commented):
LGTM. Thanks

dsneddon (changes requested):
This needs an example in the samples directory. It could be added to the existing bond examples, but it must be documented. I think we can forego unit tests for such a small change, but it needs to be documented before merging.

dsneddon:
Thank you very much for this submission. It needs a quick docs update by adding the param to the bond example json/yaml files.

os_net_config/impl_nmstate.py
```diff
@@ -152,7 +152,7 @@ def parse_bonding_options(bond_options_str):
 
 
 def set_linux_bonding_options(bond_options, primary_iface=None):
-    linux_bond_options = ["updelay", "miimon", "lacp_rate"]
+    linux_bond_options = ["updelay", "miimon", "lacp_rate", "fail_over_mac"]
```
vcandapp, line 155: Since we allow only this subset of bond options, can a warning be added if user providers other bond options like: downdelay, lacp_rate, balance_slb, etc ?

dsneddon: More importantly can we add support for additional Linux bond options? downdelay and lacp_rate should be supported. In fact we should support the whole array of options that are supported in the nmstate schema.

SharonBX: i will add to the linux_bond_options list all supported options. (btw options which are not mode that not include in the list will be ignored ... so 'downdelay' is currently not supported)

karthiksundaravel:
Abandoning it. New PR is raised https://github.com/os-net-config/os-net-config/pull/118

PR #122, opened 2024-11-07
avoid KeyError: 'drivers_autoprobe' when use os-net-config-sriov

SharonBX:
When sriov_config service is run after updating the os-net-config
package in an existing deployment, we get the error “KeyError:
‘drivers_autoprobe’“. This is due to the previously configured
sriov_config.yaml which does not have the `drivers_autoprobe` field.
Adding a check for the existence of the ’drivers_autoprobe` field to
avoid the error

### DIFF
--- a/os_net_config/sriov_config.py
+++ b/os_net_config/sriov_config.py
@@ -417,8 +417,9 @@ def configure_sriov_pf(execution_from_cli=False, restart_openvswitch=False):
             # It has to happen before we set_numvfs
             if vdpa and is_mlnx:
                 configure_switchdev(item['name'])
-            set_drivers_autoprobe(item['name'], item['drivers_autoprobe'])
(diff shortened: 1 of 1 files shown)


dsneddon:
I’d like to approve this, but I still don’t see a description for the pull request. Sharon dan you add that, or perhaps cancel the pull request and create a new one with a description of the change. Either way I like the change and will approve if there is a description for the pull request.

SharonBX:
description added yesterday
https://github.com/os-net-config/os-net-config/pull/122/commits/06b92c7b1fb7b2aee35f9b4319a4465e9905c9fe
should i add description in other places ?

karthiksundaravel (approved):
LGTM. Thanks

raukadah:
recheck after https://github.com/openstack-k8s-operators/ci-framework/pull/2526 got merged

dsneddon (approved):
Looks good, thanks!

PR #123, opened 2024-11-07
cli: make ifcfg as the default provider

karthiksundaravel:
If the host supports, ifcfg provider make that as the default provider.


karthiksundaravel:
Abandoning it, since the selection of provider is always decided by edpm-ansible.

PR #124, opened 2024-11-08
ifcfg: Fix linux bridge

Akrog:
When trying to create a Linux Bridge with something like

```
- type: linux_bridge
  name: br-ex
  mtu: 1500
  use_dhcp: false
  dns_servers: ['192.168.140.80']
  domain: ['ctlplane.example.com', 'internalapi.example.com', 'storage.example.com', 'tenant.example.com']
  addresses:
  - ip_netmask: 192.168.140.100/24
  routes: [{'destination': '0.0.0.0/0', 'nexthop': '192.168.140.1'}]
```

We get and exception saying that `route6_path` variable was used before assignment.

That happens because the variable is actually called `br_route6_path`, and this patch fixes this.


karthiksundaravel (approved):
LGTM. Thanks

PR #126, opened 2024-11-15
sriov: Apply the PF/VF config only when there is a change

karthiksundaravel:
The desired and the current states of the devices are compared for PFs and VFs. The desired state is applied only when there is a change in configuration.


dsneddon (approved):
Thanks for making the requested changes to the original patch, this looks good to me.

rjarry (approved):

dsneddon (approved):

PR #129, opened 2024-11-20
adding the option to add route though next-hop-interface

SharonBX:
nmstate support to add route to interface through local device ( ip r add default dev ). 
Not defining next-hop-address is equal to next-hop-address: 0.0.0.0

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -992,9 +992,15 @@ class NmstateNetConfig(os_net_config.NetConfig):
             if route.ip_netmask:
                 route_data[NMRoute.DESTINATION] = route.ip_netmask
             if route.next_hop:
-                route_data[NMRoute.NEXT_HOP_ADDRESS] = route.next_hop
-                route_data[NMRoute.NEXT_HOP_INTERFACE] = interface_name
+                if route.next_hop == "self":
+                    route_data[NMRoute.NEXT_HOP_INTERFACE] = interface_name
+                else:
+                    route_data[NMRoute.NEXT_HOP_ADDRESS] = route.next_hop
+                    route_data[NMRoute.NEXT_HOP_INTERFACE] = interface_name
                 if route.default:
+                    if route.next_hop == "self":
+                        msg = "self as next_hop allowed only with ip_netmask"
+                        raise os_net_config.ConfigurationError(msg)
                     if ":" in route.next_hop:
                         route_data[NMRoute.DESTINATION] = \
                             IPV6_DEFAULT_GATEWAY_DESTINATION
--- a/os_net_config/schema.yaml
+++ b/os_net_config/schema.yaml
@@ -121,7 +121,9 @@ definitions:
         type: string
         pattern: >-
           ^(?=^.{1,255}$)(?!.*\.\..*)(.{1,63}\.)+(.{0,63}\.?)|(?!\.)(?!.*\.\..*)(^.{1,63}$)|(^\.$)$
-
+    self_interface:
+        type: string
+        pattern: "^self$"
     list_of_domain_name_string_or_domain_name_string:
         oneOf:
           - type: array
@@ -169,7 +171,9 @@ definitions:
         oneOf:
         - properties:
             next_hop:
-                $ref: "#/definitions/ip_address_string_or_param"
+              oneOf:
+                - $ref: "#/definitions/ip_address_string_or_param"
+                - $ref: "#/definitions/self_interface"
             ip_netmask:
                 $ref: "#/definitions/ip_cidr_string_or_param"
             default:
--- a/os_net_config/tests/test_impl_nmstate.py
+++ b/os_net_config/tests/test_impl_nmstate.py
@@ -656,6 +656,46 @@ class TestNmstateNetConfig(base.TestCase):
         self.assertEqual(yaml.safe_load(expected_route_table),
                          self.get_route_config('em1'))
 
+    def test_network_with_route_via_device(self):
+        expected_route_table = """
+               - destination: 0.0.0.0/0
+                 metric: 10
+                 next-hop-interface: em1
+               - destination: 172.22.0.0/24
+                 metric: 100
+                 next-hop-address: 172.20.0.1
+                 next-hop-interface: em1
+           """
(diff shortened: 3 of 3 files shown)


karthiksundaravel (changes requested):
Add description to the commit message. Please add unit test cases as well

dsneddon (changes requested):
I don't think the next_hop can ever be the name of the interface as suggested in the code.

os_net_config/impl_nmstate.py
```diff
@@ -992,7 +992,8 @@ def _add_routes(self, interface_name, routes=[]):
             if route.ip_netmask:
                 route_data[NMRoute.DESTINATION] = route.ip_netmask
             if route.next_hop:
-                route_data[NMRoute.NEXT_HOP_ADDRESS] = route.next_hop
+                if route.next_hop != interface_name:
```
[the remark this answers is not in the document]
SharonBX: Am I missing something ?

    route:
        type: object
        oneOf:
        - properties:
            next_hop:
                $ref: "#/definitions/ip_address_string_or_param"
                
    ip_address_string_or_param:
        oneOf:
          - $ref: "#/definitions/ip_address_string"
          - $ref: "#/definitions/param"

karthiksundaravel:
recheck

dsneddon:
> I don't think the next_hop can ever be the name of the interface as suggested in the code. 

After looking more closely at the schema.yaml, the next_hop can be a string. I suggest allowing "self" as this would align with "next-hop-self" configuration in network hardware, and would work with mapped NICs. We would need to use "ip_netmask: 0.0.0.0/0" or "ip_netmask: ::/0" and not "default: true" in order to tell whether a route is v4 or v6, and "next_hop: self" should set the next_hop_interface to the interface name for nmstate.

dsneddon (changes requested):
Please allow "self" for the next_hop to set the next_hop_interface, and we'll have to use ip_netmask with 0.0.0.0/0 or ::/0 to determine if the route is v4 or v6 when using default route.

os_net_config/impl_nmstate.py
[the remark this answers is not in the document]
dsneddon: No you are correct

dsneddon:
> I don't think the next_hop can ever be the name of the interface as suggested in the code.

This turned out to not be fully true, we do allow strings as get_param or get_input, but if we are accepting raw strings we should definitely filter our inputs (the latest patchset does this in schema.yaml to allow “self” specifically).

dsneddon:
Recheck

dsneddon (approved):
This is a good start, thanks. I think we need to expand on this, but LGTM.

os_net_config/impl_nmstate.py
```diff
@@ -992,9 +992,15 @@ def _add_routes(self, interface_name, routes=[]):
             if route.ip_netmask:
                 route_data[NMRoute.DESTINATION] = route.ip_netmask
             if route.next_hop:
-                route_data[NMRoute.NEXT_HOP_ADDRESS] = route.next_hop
-                route_data[NMRoute.NEXT_HOP_INTERFACE] = interface_name
+                if route.next_hop in "self":
```
karthiksundaravel, line 995: can we use == ?

os_net_config/impl_nmstate.py
```diff
@@ -992,9 +992,15 @@ def _add_routes(self, interface_name, routes=[]):
...
+                    route_data[NMRoute.NEXT_HOP_INTERFACE] = interface_name
+                else:
+                    route_data[NMRoute.NEXT_HOP_ADDRESS] = route.next_hop
+                    route_data[NMRoute.NEXT_HOP_INTERFACE] = interface_name
                 if route.default:
+                    if "self" in route.next_hop:
```
karthiksundaravel, line 1001: same as above

karthiksundaravel (changes requested):

os_net_config/schema.yaml
```diff
@@ -15,6 +15,8 @@ definitions:
             get_input:
                 type: string
           additionalProperties: False
+        - type: string
+          pattern: "^self$"
```
vcandapp, line 19: I have reservation on changing the base param - because base param will be used by:
- dns_server ( "ip_address_string_or_param" )
- ip_cidr_string_or_param
- route_rule
Would this change make "self" being allowed for all of the above ?

Maybe a new custom type is safer

vcandapp (changes requested):

os_net_config/schema.yaml
SharonBX: every $ref: "#/definitions/param call will allow "self" (as is) string.
do you see any risk here ?

dsneddon: There is no risk with accepting self as-is, the danger is if we open this up in a follow-up to accept interface names, but I think it's acceptable, after all the user running this as root is also the one supplying the input template, so they accept the risk themselves.

dsneddon: Viji is correct about not changing this base type but rather creating a custom type, we don't want to limit route rules or dns server to only being "self", so we want to have a custom type here that applies to the route object but not route_rule or dns_server, etc.

dsneddon: I think we should revert to allowing all strings for now, as I mentioned the risk is the same as we had before, and the user running os-net-config is providing the input so they are accepting the risk. Perhaps we can come up with a more restrictive string that doesn't allow escape characters, but I think that the oslo process execution module we are using already has some input sanitizing, so I think we are covered. Let's just allow any string for now and mark this for review in the near future.

SharonBX: OK
added patchset (to avoid scheme errors) with new type "self_interface" with self string for now ... you can expend the regex to accept interface names later on for

karthiksundaravel (approved):
LGTM!

vcandapp:
@SharonBX , can you please add appropriate resolution to all open review comments and close them - so that I can merge this PR ?

SharonBX:
> Add description to the commit message. Please add unit test cases as well

done

SharonBX:
> Add description to the commit message. Please add unit test cases as well

done

PR #131, opened 2024-11-20
ignore VFs defined 

SharonBX:
currently os-net-config fail to rerun in case config.json not includes sriov VF definitions (or sriov been defined on the host without nmstate ) the fail reason is the desire VFs hard coded to zero is not as current.
this pr will replace the zero with current defined VFs to avoid the fail verification.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -45,6 +45,7 @@ from os_net_config import common
 from os_net_config import objects
 from os_net_config import utils
 
+
 logger = logging.getLogger(__name__)
 
 # Import the raw NetConfig object so we can call its methods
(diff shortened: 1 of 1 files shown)


os_net_config/impl_nmstate.py
```diff
@@ -43,8 +43,10 @@
 import os_net_config
 from os_net_config import common
 from os_net_config import objects
+from os_net_config import sriov_config
```
karthiksundaravel, line 46: Don't import sriov_config here. The numvfs could be obtained from the iface_state() itself.

karthiksundaravel (changes requested):
Its tricky one and I'll have to check how this patch fits the other Openstack use cases.

dsneddon (approved):

dsneddon (approved):

dsneddon (approved):
This looks good to me, but I will wait for a review from Karthik before merging. Thanks!

karthiksundaravel (approved):

PR #134, opened 2024-12-04
ovs: keep the ovs internal interface name same as the bridge

karthiksundaravel:
The ovs internal interface is named as {bridge_name}-p earlier. The name is changed to {bridge_name} now so that it is consistent with the ifcfg provider.


dsneddon (approved):

vcandapp (approved):
lgtm

PR #137, opened 2024-12-18
[DNM]Ignore defined ip rules

SharonBX:
in case ip rules defined not by NetworkManager, os net will fail to remove them 
and will endup with Verification Error. 
with this pr os-net-config will clean or ignore existing ip rules by cleanup argument.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -356,6 +356,7 @@ class NmstateNetConfig(os_net_config.NetConfig):
         # separately if the flag is set. It will be applicable
         # only for NIC Partitioning use cases.
         self.need_vf_config = False
+        self.cleanup = False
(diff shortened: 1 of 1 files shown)


dsneddon (approved):

karthiksundaravel (changes requested):
Sorry Sharon, this approach will lead to confusion in the handling of ip rules. The usage of cleanup variable to remove the undesired `ip rules`  would affect the normal functionality. We need to identify a mechanism where we can allow the ip rules configured by other tools (like wireshark) to co-exist with that of those defined by os-net-config. With ifcfg provider we do have a clear differentiation due to the choice of rule-file used. We need to identify the possibility of having something similar with nmstate.

karthiksundaravel:
New PR [1] is raised for the same and hence abandoning it.
[1] https://github.com/os-net-config/os-net-config/pull/159

PR #139, opened 2024-12-24
utils: simplify get_pci_address()

karthiksundaravel:
Presently the PCI address is read using ethtool command, when the device is bound with standard netdev driver. It is fetched from dpdk map when its bound with vfio-pci. The caller is expected to identify the device type and choose the appropriate function. The get_pci_address(interface_name) will now handle fetching the PCI address of the interface irrespective of the device types. In addition to that, now the PCI address of the SR-IOV VFs also could be fetched by passing the interface name in the format f"{pf_name}:{vfid}".

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -84,6 +84,10 @@ class OvsDpdkBindException(ValueError):
     pass
 
 
+class SriovVfNotFoundException(ValueError):
+    pass
+
+
 def set_noop(value):
     global noop
     noop = value
@@ -143,10 +147,8 @@ def get_dev_path(ifname, path=None):
     return os.path.join(SYS_CLASS_NET, ifname, path)
 
 
-def get_pci_dev_path(pci_address, path=None):
-    if not path:
-        path = ""
-    elif path.startswith("_"):
+def get_pci_dev_path(pci_address, path=""):
+    if path.startswith("_"):
         path = path[1:]
     return os.path.join(_SYS_BUS_PCI_DEV, pci_address, path)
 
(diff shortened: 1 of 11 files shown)


karthiksundaravel:
recheck

os_net_config/impl_nmstate.py
```diff
@@ -45,7 +45,6 @@
 from os_net_config import objects
 from os_net_config import utils
 
-
```
dsneddon, line 48: It's not unusual to have two white space lines after includes, but I'm not attached to either way.

os_net_config/tests/test_cli.py
```diff
@@ -75,17 +75,15 @@ def run_cli(self, argstr, exitcodes=(0,)):
...
-            return "0000:00:07.0"
-        if 'eth1' in ifname:
-            return "0000:00:08.0"
-        if 'eth2' in ifname:
-            return "0000:00:09.0"
-        if 'em3' in ifname:
-            return "0000:00:03.0"
-        if 'em1' in ifname:
-            return "0000:00:01.0"
+    def stub_get_pci_address(self, ifname):
```
dsneddon, line 78: This map is an improvement over the list of if statements. Nice work.

dsneddon (commented):
I have a few thoughts. First of all I think we might want an extra log statement or two if debug is set since we won't get the output from ethtool. This change is fairly extensive for the amount of change, but I'm not sure it can be broken up into smaller patchsets. I'll discuss logging and comments before deciding if I approve this without more work.

os_net_config/common.py
```diff
@@ -324,6 +326,60 @@ def is_vf(pci_address):
     return is_sriov_vf
 
 
+def get_pci_address(name):
+    """Fetch the PCI address of the interface
```
dsneddon, line 330: I believe ethtool gets this info from the same place, so the results should be the same and it should be much faster to make a filesystem call than making an external process call to ethtool.

dsneddon (approved):
It's too bad this change required so many files and lines of code to be updated, but the actual scope of the change isn't very large. We could have replaced the ethtool process call in a separate PR, but that wouldn't have reduced the size of the patch very much.

os_net_config/utils.py
```diff
@@ -377,11 +332,11 @@ def get_dpdk_devargs(ifname, noop):
...
                 # It includes the VFs as well. For all Other Mellanox devices
                 # the PCI address are not stored in dpdk_mapping.yaml file,
-                # so we need to get their pci address with ethtool.
-                dpdk_devargs = get_pci_address(ifname, noop)
+                # so we need to get their pci address from sysfs.
+                dpdk_devargs = common.get_pci_address(ifname)
         else:
             logger.info("%s: Getting stored PCI address as devarg", ifname)
-            dpdk_devargs = get_stored_pci_address(ifname, noop)
+            dpdk_devargs = common.get_pci_address(ifname)
```
vcandapp, line 339: Log "etting stored PCI address as devarg" -> it means devargs for non - mellanox nic ???

os_net_config/utils.py
```diff
@@ -221,7 +217,7 @@ def diff(filename, data):
 
 def update_dpdk_map(ifname, driver):
     noop = common.get_noop()
-    pci_address = get_pci_address(ifname, noop)
+    pci_address = common.get_pci_address(ifname)
```
vcandapp, line 220: noop line can be removed ?

os_net_config/utils.py
karthiksundaravel: I'll update the logs so its clear. Only for CX3 cards the dpdk devargs is 
    _dpdk_devargs = f"class=eth,mac={common.interface_mac(ifname)}"_
For all others devices its the pci address. Now we have made a single entry for fetching the pci address via get_pci_address(), so we can simplify this function better as well. Possibly in a follow up patch.

karthiksundaravel:
recheck

karthiksundaravel:
recheck

karthiksundaravel:
recheck

os_net_config/common.py
```diff
@@ -143,10 +147,8 @@ def get_dev_path(ifname, path=None):
     return os.path.join(SYS_CLASS_NET, ifname, path)
 
 
-def get_pci_dev_path(pci_address, path=None):
-    if not path:
-        path = ""
-    elif path.startswith("_"):
+def get_pci_dev_path(pci_address, path=""):
+    if path.startswith("_"):
```
rjarry, line 151: I am not sure what this change is for?

os_net_config/common.py
```diff
@@ -324,6 +326,61 @@ def is_vf(pci_address):
...
+    device = name.split(":")
+    ifname = device[0]
+    if len(device) == 3 and ifname == "sriov":
+        ifname = device[1]
+        vfid = device[2]
+
+    if vfid:
+        dev_path = get_dev_path(ifname, f'virtfn{vfid}')
+    else:
+        dev_path = get_dev_path(ifname, '_device')
```
rjarry, line 350: This is a bit obscure. Maybe we could have an explicit way to determine which path we get? Instead of relying on "_" at the beginning of the argument.

PR #143, opened 2024-12-29
Pf given to guest

karthiksundaravel:
sriov: Skip PF configuration when its attached to guest
  
In scenarios where the PF is attached to the guest, the PF device will not be visible in the hosts net subsystem. And if os-net-config is rerun in that scenario, the NIC mapping would fail. Checks shall be added so that the situation where the PF is attached to the guest is detected and handled.


karthiksundaravel:
recheck

rjarry (approved):

PR #144, opened 2024-12-29
Driver enhacement

karthiksundaravel:
pci: fetch the driver from pci address
    
The existing implementation of get_driver(), get_interface_driver()
uses ethtool or parses uevent file from netdev's sysfs. This may not
be reliable when the devices are bound to drivers like vfio-pci or not
bound with any drivers. Now, the implementation requires fetching the
the pci address of the device and then perform the below command to
fetch the driver.
basename $(readlink /sys/bus/pci/devices/{pci_address}/driver)


rjarry (dismissed):

rjarry (approved):

dsneddon (approved):
It seems that all my issues were addressed in the latest patchset except for one obsolete comment, no reason not to get this merged.

PR #142, opened 2024-12-26
sriov: add pci address and mac address in sriov map

karthiksundaravel:
sriov: add pci address and mac address in sriov map
SR-IOV PF's pci address and mac address shall be stored
in the map, so that when the PF is attached to the guest, the
PF's details coud be read.

Signed-off-by: Karthik Sundaravel <ksundara@redhat.com>

PR #145, opened 2024-12-30
sriov: Handle NIC partitioning while autoprobe is disabled

karthiksundaravel:
sriov: Handle NIC partitioning while autoprobe is disabled
    
When drivers_autoprobe is disabled for sriov_pf, the VFs created
will not be bound with the required net drivers and hence those
VFs would not be listed in the /sys/class/net/<vf name>.
In order to make the NIC partitioning functional the required drivers
needs to be bounded with the VF.

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -434,6 +434,25 @@ def is_vf_by_name(interface_name, check_mapping_file=False):
     return is_sriov_vf
 
 
+def get_default_vf_driver(pf_name, vfid):
+    modalias_path = get_dev_path(pf_name, f"virtfn{vfid}/modalias")
+    try:
+        with open(modalias_path) as f:
+            alias = f.read().strip()
+        cmd = ["modprobe", "-R", alias]
+        out, err = processutils.execute(*cmd)
+        kernel_driver = out.strip()
+        logger.info(
+            "%s-%d: default vf driver is %s", pf_name, vfid, kernel_driver
+        )
+        return kernel_driver
+    except (OSError, processutils.ProcessExecutionError) as e:
+        logger.error(
+            "%s-%d: failed to get default vf driver: %s", pf_name, vfid, e
+        )
+        return None
+
+
 def set_driverctl_override(pci_address, driver):
     if driver is None:
         logger.info("%s: Driver override is not required.", pci_address)
--- a/os_net_config/objects.py
+++ b/os_net_config/objects.py
@@ -23,6 +23,7 @@ import logging
 import netaddr
 from oslo_utils import strutils
 
+import os_net_config
 from os_net_config import common
 from os_net_config import utils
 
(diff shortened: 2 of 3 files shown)


rjarry (approved):
rebase on top of #142 and #144 and LGTM :)

rjarry (changes requested):

os_net_config/objects.py
```diff
@@ -1543,6 +1543,18 @@ def __init__(self, device, vfid, use_dhcp=False, use_dhcpv6=False,
...
+        # the VFs will not be bound with the appropriate netdev drivers.
+        # When VFs are used for host networking (NIC Partitioning), then the
+        # VF shall be bound with the default net driver, so that the VF's
+        # name and other details could be fetched and configured. If the VF
+        # is attached to a DPDK port, then the below override would be overridden
+        # again later. 
+        # When autoprobe is enabled, the current driver and the default net driver
+        # will be the same, and hence no override happens.
+        driver = common.get_default_vf_driver(device, vfid)
+        common.set_driverctl_override(pci_address, driver)
```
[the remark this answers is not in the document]
dsneddon: That's a good idea

os_net_config/utils.py
```diff
@@ -564,20 +564,23 @@ def configure_sriov_vfs():
     nicpart_udev_rules_check()
 
 
-def get_vf_devname(pf_name, vfid):
+def get_vf_devname(pf_name, vfid):
```
rjarry, line 567: remove trailing whitespace

rjarry (changes requested):

os_net_config/common.py
```diff
@@ -434,6 +434,25 @@ def is_vf_by_name(interface_name, check_mapping_file=False):
     return is_sriov_vf
 
 
+def get_default_vf_driver(pf_name, vfid):
+    modalias_path = get_dev_path(pf_name, f"virtfn{vfid}/modalias")
```
vcandapp, line 438: Will noop() handling be needed here ?

os_net_config/objects.py
```diff
@@ -1589,9 +1590,29 @@ def __init__(self, device, vfid, use_dhcp=False, use_dhcpv6=False,
...
+        # If the VF is attached to a DPDK port, then the below driverctl
+        # override would be overridden again later.
+        # When autoprobe is enabled, the default driver would be already bound
+        # with the VF, and hence no override happens.
+        # During update or os-net-config re-run, the device could be already
+        # bound with vfio-pci and the driver overriding is avoided.
+        def_driver = common.get_default_vf_driver(device, vfid)
+        cur_driver = common.get_pci_device_driver(pci_address)
+        if cur_driver is None:
+            common.set_driverctl_override(pci_address, def_driver)
```
vcandapp, line 1612: There is double call made to "get_pci_device_driver()" - one in L1610 and another inside set_driverctl_override(). Is there scope to optimize ?

vcandapp:
@karthiksundaravel , the Jira link to PR is missing

os_net_config/common.py
karthiksundaravel: Its not needed. Its read only function and the unit tests shall stub this when required.

vcandapp: Will this path exist for noop case ? Like  get_pci_device_driver() https://github.com/os-net-config/os-net-config/blob/master/os_net_config/common.py#L313 uses noop check - as the pci/sysfs will not exist right ?

karthiksundaravel: noop is actually meant for dry run on the real hardware, so all the read actions should be acceptable. however the set_driverctl_override() will break the noop functionality. Lets handle this in a follow up PR as its unrelated to this.

vcandapp (approved):
LGTM

rjarry (approved):

PR #147, opened 2025-01-10
DCB configurations over OVS DPDK BOND

abhiramnarayana:
For the DCB configurations over a dpdkbond, currently the configs needs to be repeated under all the dpdk bond members (interfaces). This patch adds support in os-net-config to be able to apply DCB configurations directly under the object of type ovs dpdk bond.

(cherry picked from commit 955a57cc637988c63f0f41565bfc2c3c29a30efc)
Signed-off-by: Abhiram R N <abhiramrn@gmail.com>

### DIFF
--- a/os_net_config/objects.py
+++ b/os_net_config/objects.py
@@ -1757,6 +1757,39 @@ class OvsDpdkBond(_BaseOpts):
             self.primary_interface_name = bond_members[0].name
 
     @staticmethod
+    def update_port_dcb_config(members, dcb_config_json, mapped_nic_names):
+        for ovsdpdkport in members:
+            ovsdpdkport_json = ovsdpdkport.members
+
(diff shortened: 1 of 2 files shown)


karthiksundaravel (approved):
lgtm. Thanks

dsneddon (changes requested):
This will need unit tests and we should test cases where DCB is set on bond, members, and both. Also, should we possibly warn or fail if DCB is different between bond and members (if it is set in multiple places)?

abhiramnarayana:
Hi @dsneddon , 
Since this is a backport PR. Like Karthik suggested its always best we fix it in master and then bring to wallaby.
Hence kindly request to approve and help merge.

CC: @karthiksundaravel @vcandapp 

Thanks & Regards,
Abhiram R N

PR #149, opened 2025-01-10
noop: Run os-net-config without applying the changes

abhiramnarayana:
When os-net-config is run with noop, its failing due to permission issues or trying to write files and configure devices. When `noop` is used, there shall be no changes made to the machine, while the configuration files or templates derived shall be output on the console. This is done so that the template/compatibility issues of the config.yaml could be validated on a given hardware without changing the system files or configuration.

(cherry picked from commit 2e60c77aee02d682028ff7795f5df1e9c9a0ae4e)
Signed-off-by: Abhiram R N <abhiramrn@gmail.com>

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -178,6 +178,9 @@ def load_provider(name, noop, root_dir):
 
 def main(argv=sys.argv, main_logger=None):
     opts = parse_opts(argv)
+
+    common.set_noop(opts.noop)
+
     if not main_logger:
         main_logger = common.configure_logger(log_file=not opts.noop)
     common.logger_level(main_logger, opts.verbose, opts.debug)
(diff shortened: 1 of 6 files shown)


dsneddon (commented):
Please comment to indicate if this has been tested with --noop with both ifcfg and nmstate providers. If it has been tested with both providers against a variety of the sample configs in etc/os-net-config/samples then I will approve.

dsneddon (approved):

dsneddon:
Sorry, I didn't mean to approve, Karthik found an error with SR-IOV

dsneddon:
Apparently the error is in the original change in master as well. It might be easiest to merge this change and create a follow up patch with a unit test that triggers the error and a fix, merge it to master and back port.

karthiksundaravel (approved):
lgtm.

PR #150, opened 2025-01-12
noop: fix noop run for nic partitioning use case

karthiksundaravel:
The noop dry run attempts to write to the sriov map, which results in failure for non privileged users. The fix is to avoid writing to the map files during dry run.

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -183,6 +183,9 @@ def get_file_data(filename):
 
 
 def write_yaml_config(filepath, data):
+    if get_noop():
+        logger.info("Writing file %s with content %s", filepath, data)
+        return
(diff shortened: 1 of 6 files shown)


karthiksundaravel:
recheck

os_net_config/common.py
```diff
@@ -183,6 +183,9 @@ def get_file_data(filename):
 
 
 def write_yaml_config(filepath, data):
+    if get_noop():
+        logger.info("Writing %s", filepath)
```
vcandapp, line 187: Can we add "data" also to the log (instead of file path?)

vcandapp (approved):
lgtm

dsneddon:
recheck

karthiksundaravel:
recheck

vcandapp (approved):
lgtm

PR #154, opened 2025-01-14
iprules: ignore defined ip rules

SharonBX:
If the IP rules are not defined by NetworkManager, os-net-config will be
unable to remove them, resulting in a Verification Error. With this pull
request, os-net-config will record managed iprule into filesystem and
will ignore iprules not in his list.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -2012,7 +2012,7 @@ class NmstateNetConfig(os_net_config.NetConfig):
         if isinstance(ovs_dpdk_port.members[0], objects.SriovVF):
             # in case of VFs the DPDK driver will be bound using
             # dispatcher script
-            pci_address = utils.get_dpdk_pci_address(ifname)
+            pci_address = ovs_dpdk_port.members[0].pci_address
             utils.update_dpdk_map(ifname,
                                   ovs_dpdk_port.driver)
         else:
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -361,7 +361,7 @@ def main(argv=sys.argv, main_logger=None):
         for iface_json in iface_array:
             try:
                 obj = objects.object_from_json(iface_json)
-            except utils.SriovVfNotFoundException:
+            except common.SriovVfNotFoundException:
                 continue
 
             if _is_sriovpf_obj_found(obj):
@@ -410,7 +410,7 @@ def main(argv=sys.argv, main_logger=None):
             # SriovVfNotFoundException shall be raised if not available.
             try:
                 obj = objects.object_from_json(iface_json)
-            except utils.SriovVfNotFoundException:
+            except common.SriovVfNotFoundException:
                 if not opts.noop:
                     raise
 
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -84,6 +84,10 @@ class OvsDpdkBindException(ValueError):
     pass
 
 
+class SriovVfNotFoundException(ValueError):
+    pass
+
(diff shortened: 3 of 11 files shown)


os_net_config/impl_nmstate.py
```diff
@@ -98,7 +100,7 @@
 
 POST_ACTIVATION = 'post-activation'
 DISPATCH = 'dispatch'
-
+MNG_RULES = '/var/lib/os-net-config/rules.yaml'
```
karthiksundaravel, line 103: Please keep it in /var/lib/os-net-config/nmstate_files/rules.yaml

os_net_config/impl_nmstate.py
```diff
@@ -850,9 +867,11 @@ def generate_rules(self):
...
         if clear_rules:
             for c_rule in curr_rules:
-                c_rule[NMRouteRule.STATE] = NMRouteRule.STATE_ABSENT
-                del_rules.append(c_rule)
-                logger.info(f'Prepare to remove rule - {c_rule}')
+                if self.managed_iprules(c_rule):
+                    c_rule[NMRouteRule.STATE] = NMRouteRule.STATE_ABSENT
+                    del_rules.append(c_rule)
+                    logger.info(f'Prepare to remove rule - {c_rule}')
+        self.write_config(MNG_RULES, add_rules)
```
karthiksundaravel, line 874: Keep a copy of the file content and write only after apply is successful. In case of failures the initial content shall be updated in the file.

karthiksundaravel: Also the rules shall be written in nmstate template format.

os_net_config/impl_nmstate.py
```diff
@@ -828,6 +830,21 @@ def generate_routes(self, interface_name):
...
 
+    def managed_iprules(self, c_rule):
+        """
+        check if ip rule is manage by os-net.
+        only managed ip rule can be deleted, others will be ignored
+        """
+        if os.path.exists(MNG_RULES):
+            mng_iprules = ast.literal_eval(common.get_file_data(MNG_RULES))
+            for iprule in mng_iprules:
+                if c_rule['ip-from'] == iprule['ip-from'] and \
```
karthiksundaravel, line 841: Perform a strict match of all the options. Only options like metric shall be ignored

karthiksundaravel (changes requested):

os_net_config/impl_nmstate.py
```diff
@@ -2293,6 +2315,8 @@ def apply(self, cleanup=False, activate=True, config_rules_dns=True):
                 if activate:
                     self.nmstate_apply(apply_data, verify=True)
 
+            RULES_TEMPLATE["route-rules"]["config"] = add_rules
```
karthiksundaravel, line 2318: Can this be moved when all the apply() is succesfull. i.e in #Line 2332 ?

os_net_config/impl_nmstate.py
```diff
@@ -828,6 +834,21 @@ def generate_routes(self, interface_name):
                 logger.info(f'Prepare to remove route - {c_route}')
         return add_routes, del_routes
 
+    def managed_iprules(self, c_rule):
```
karthiksundaravel, line 837: rename to _is_managed_iprules() ?

os_net_config/impl_nmstate.py
karthiksundaravel: Also we don't need a global scope variable here isn't it ?

karthiksundaravel (changes requested):

PR #161, opened 2025-01-17
ovs: keep the ovs internal interface name same as the bridge

abhiramnarayana:
The ovs internal interface is named as {bridge_name}-p earlier. The name is changed to {bridge_name} now so that it is consistent with the ifcfg provider.

(cherry picked from commit 407f317564e6b72c64e35e47f4caddb138ca183f)


karthiksundaravel (approved):

vcandapp (approved):
LGTM

PR #162, opened 2025-01-17
sriov: retain the VF count for interfaces

abhiramnarayana:
Retain the sriov numvfs (if any) configured on the ethernet interface as it is. This is done so that the numvfs could be managed externally or separately in 2 different config yamls.

(cherry picked from commit 22b9a6570649d2c627e6a0f3d892ad13ce88789f)


vcandapp (approved):
LGTM

PR #165, opened 2025-01-22
dcb: add function to verify DCBX mode config using mstconfig

abhiramnarayana:
This change introduces the 'is_dcbx_mstconfig_enabled' function, which verifies if DCBX mode is correctly enabled on Mellanox NICs using the 'mstconfig' tool. Based on this, we can check if mstconfig and mstfwreset needs to be done or not.

(cherry picked from commit ff88b85e9ec3cf191826e8c5f21b8d9f9a1aa46d)

### DIFF
--- a/os_net_config/dcb_config.py
+++ b/os_net_config/dcb_config.py
@@ -17,6 +17,7 @@
 import argparse
 import logging
 import os
+import re
 import sys
 import yaml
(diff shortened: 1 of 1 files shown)


karthiksundaravel (approved):
lgtm

PR #169, opened 2025-01-24
sriov: nic partitioning with only dpdk ports fail

karthiksundaravel:
In NIC partitioning scenario with only dpdk ports, the VF configuration is skipped since add_sriov_vf() will not be called for dpdk ports. The VF configuration for dpdk ports that uses NIC partitioning are added to the sriov_vf_data.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -2015,6 +2015,10 @@ class NmstateNetConfig(os_net_config.NetConfig):
             pci_address = ovs_dpdk_port.members[0].pci_address
             utils.update_dpdk_map(ifname,
                                   ovs_dpdk_port.driver)
+            # In NIC partitioning scenario with dpdk ports, the VF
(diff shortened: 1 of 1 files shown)


os_net_config/impl_nmstate.py
```diff
@@ -2178,6 +2182,15 @@ def add_sriov_pf(self, sriov_pf):
         self.need_pf_config = True
         self.__dump_config(data, msg=f"{sriov_pf.name}: Prepared config")
```
vcandapp, line 2184: Above lines 2171-2173 is duplicate of line 2166-2167 (setting promisc)

vcandapp (changes requested):

karthiksundaravel:
recheck

karthiksundaravel:
recheck

karthiksundaravel:
recheck

karthiksundaravel:
recheck

vcandapp (approved):
lgtm

PR #175, opened 2025-01-27
Cleanup nmconnections

rabi:
Sometimes there are NM configured interfaces that needs cleanup for them to not conflict.


os_net_config/impl_ifcfg.py
```diff
@@ -1984,6 +1987,11 @@ def apply(self, cleanup=False, activate=True, config_rules_dns=True):
                         )
                         self.ifdown(interface_name)
                         self.remove_config(ifcfg_file)
+            for nmconn_file in glob.iglob(nm_cleanup_pattern()):
+                interface_name = Path(nmconn_file).stem
+                cmd = ['nmcli', 'connection', 'delete', interface_name]
```
karthiksundaravel, line 1992: instead of `nmcli` can we use nmstate ?

rabi: You mean use nmstate api or `nmstatectl`? We do use `nmcli`[1] elsewhere right? 

[1] https://github.com/os-net-config/os-net-config/blob/master/os_net_config/impl_nmstate.py#L372

os_net_config/impl_ifcfg.py
```diff
@@ -1984,6 +1987,11 @@ def apply(self, cleanup=False, activate=True, config_rules_dns=True):
...
+                self.execute(msg, *cmd)
```
abhiramnarayana, line 1993: It will be good to keep this self.execute inside of a try except block so that in case of failure its doesnt silently fail.

abhiramnarayana, line 1993: 'msg' seems to be not assigned. can you check?

rabi: Not sure what you mean by fail silently. It would be the same as ifdown() call in L1985 which also uses cli command. Unless you want to log something here or ignore the error I don't see why one would use try/except block.

rabi: good catch, thanks!

rabi:
recheck tempest failure

arxcruz (approved):

karelyatin:
recheck fresh results

rabi:
recheck

rabi:
@karthiksundaravel Can we merge this or you still have concerns?

aslvbnnm:
I can confirm that manually deleting the nm connection created by cloud-init during the run of edpm-svc configure network (which runs os-net-config) solves the connection hangup issue on the ctlplane vlan.
The issue was observed in 
 os-net-config-18.0.1-0.20250114153726.d268100.el9.noarch

karthiksundaravel:
@aslvbnnm Is the device configured by cloud-init also part of the config.yaml ? If yes what is the difference in configuration ? Can you please share us the details of the nm connection (ifcfg or key file) created by cloud init ?

rabi:
> @aslvbnnm Is the device configured by cloud-init also part of the config.yaml ?

@karthiksundaravel Not sure which config.yaml you're talking about, the one generated for os-net-config? Please check the linked jira for the details, the devices are created during baremetal provsioning.  There is always a possibility that nodes would have some interfaces configured that's not in cofig data for os-net-config. We have option to clean them or leave after os-net-config runs.

rabi:
> @aslvbnnm Is the device configured by cloud-init also part of the config.yaml ?

@karthiksundaravel Not sure which config.yaml you're talking about, the one generated for os-net-config? Please check the linked jira for the details, the devices are created during baremetal provsioning.  There is always a possibility that nodes would have some interfaces configured that's not in cofig data for os-net-config. We have option to clean them or leave after os-net-config runs.

os_net_config/impl_ifcfg.py
Jaganathancse: @karthiksundaravel As discussed with Rabi, this clean up is required for the interfaces which are having nm connection created during baremetal provisioning.

os_net_config/impl_ifcfg.py
```diff
@@ -1984,6 +1989,12 @@ def apply(self, cleanup=False, activate=True, config_rules_dns=True):
                         )
                         self.ifdown(interface_name)
                         self.remove_config(ifcfg_file)
+            for nmconn_file in glob.iglob(nm_cleanup_pattern()):
+                interface_name = Path(nmconn_file).stem
+                cmd = ['nmcli', 'connection', 'delete', interface_name]
+                msg = "Removing nm connection %s" % interface_name
```
Jaganathancse, line 1995: Can we updae this log meesage as  similar like ifcfg message, here also removing connection and deleting nm connection file?
or 
This message is specific to nm connection removal and part of execute function, can we add above one more log for cleanup message to have similar like ifcfg cleanup?

os_net_config/impl_ifcfg.py
```diff
@@ -1984,6 +1989,12 @@ def apply(self, cleanup=False, activate=True, config_rules_dns=True):
                         )
                         self.ifdown(interface_name)
                         self.remove_config(ifcfg_file)
+            for nmconn_file in glob.iglob(nm_cleanup_pattern()):
+                interface_name = Path(nmconn_file).stem
+                cmd = ['nmcli', 'connection', 'delete', interface_name]
```
karthiksundaravel, line 1994: Can you also look for any dispatcher scripts present in `/etc/NetworkManager/dispatcher.d/<scripts>` and remove the same. There are some scripts created by nmstate in this path. May be this could be handled in a separate PR.

karthiksundaravel (changes requested):
Overall, it looks good to me. I hope we don't cleanup the connections that would be necessary for the deployments to go through.

os_net_config/impl_ifcfg.py
rabi: I don't see that format used anywhere in theimpl_ifcfg.py[1]. Is this a guideline specific to this project? If that is the reason you're blocking, I can update it though.

[1] https://github.com/os-net-config/os-net-config/blob/master/os_net_config/impl_ifcfg.py#L125

os_net_config/impl_ifcfg.py
rabi: I've not seen anything else creating issues with deployment.

rabi:
> I hope we don't cleanup the connections that would be necessary for the deployments to go through.

Why would we? We only cleanup when the user explictily sets the `cleanup: true` flag and cleanup connections not in the config. This is a feature[1] we have for ages with os-net-config.

[1] https://github.com/os-net-config/os-net-config/commit/462436fa4d339c80a888def1707f792c33cca715

rabi:
recheck test-operator timeout

karthiksundaravel (approved):
lgtm. Thanks.

os_net_config/impl_ifcfg.py
karthiksundaravel: We are converging on the same, so that the logs could be interpreted based on device.  We are yet to handle it completely.
Example: https://github.com/os-net-config/os-net-config/blob/master/os_net_config/impl_ifcfg.py#L785

PR #176, opened 2025-01-29
Merge d0b96b99f61d1366ab7955635ae8c35e432c5948 into 5ff3b7c19e8a0bc52…

vcandapp:
In NIC partitioning scenario with only dpdk ports, the VF configuration is skipped since add_sriov_vf() will not be called for dpdk ports. The VF configuration for dpdk ports that uses NIC partitioning are added to the sriov_vf_data.

(cherry-pick from master)
…80ff72d7c500d3facde51ac

(cherry picked from commit d91e52d08c56537c0970693635ff6b936dc2849c)


karthiksundaravel:
recheck

PR #178, opened 2025-02-17
ifcfg: verify file existence before removal

rabi:
Verify file existence before attempting removal to prevent errors.

jira: https://github.com/os-net-config/os-net-config/pull/178

### DIFF
--- a/os_net_config/__init__.py
+++ b/os_net_config/__init__.py
@@ -492,7 +492,7 @@ class NetConfig(object):
     def remove_config(self, filename, msg=None):
         msg = msg or "Removing config %s" % filename
         logger.info("%s%s", self.log_prefix, msg)
-        if not self.noop:
+        if not self.noop and os.path.exists(filename):
             os.remove(filename)
 
     def ifdown(self, interface, iftype='interface'):


karthiksundaravel (changes requested):
Can you please modify the commit message like
"The existence of the config file is validated before proceeding for removing the same".  
Also it will be helpful for us if you could explain the scenario where the failure happens.

rabi:
> I don't see anything wrong with the current message which says don't fail when the file does not exist as that's a valid scenario. I'm not sure how the message you suggested is going to be more helpful.

This is the error if you're interested. It's possible that some devices don't have entries in /etc/NetworkManager/system-connections.

2025-02-14 00:52:21.978 INFO os_net_config.remove_config Removing config /etc/NetworkManager/system-connections/ci-private-network.nmconnection
2025-02-14 00:52:21.978 ERROR os_net_config.main ***Failed to configure with ifcfg provider***
[Errno 2] No such file or directory: '/etc/NetworkManager/system-connections/ci-private-network.nmconnection'
2025-02-14 00:52:21.979 ERROR os_net_config.common.log_exceptions Traceback (most recent call last):
  File "/bin/os-net-config", line 10, in <module>
    sys.exit(main())
  File "/usr/lib/python3.9/site-packages/os_net_config/cli.py", line 423, in main
    files_changed = provider.apply(cleanup=opts.cleanup,
  File "/usr/lib/python3.9/site-packages/os_net_config/impl_ifcfg.py", line 1997, in apply
    self.remove_config(nmconn_file)
  File "/usr/lib/python3.9/site-packages/os_net_config/__init__.py", line 496, in remove_config
    os.remove(filename)
FileNotFoundError: [Errno 2] No such file or directory: '/etc/NetworkManager/system-connections/ci-private-network.nmconnection'
NoneType: None

karthiksundaravel:
> directory

karthiksundaravel:
Hi Rabi,
The commit message here doesn't use words like 'I', 'we', 'you' and hence asked for a message that explains the change. Also looking at the error, is there any scenario where the /etc/NetworkManager/system-connections/<connection> file still exists after the connection is deleted ? Looks like we don't need the `self.remove_config(nmconn_file)` at all.

rabi:
> Hi Rabi, The commit message here doesn't use words like 'I', 'we', 'you' and hence asked for a message that explains the change. Also looking at the error, is there any scenario where the /etc/NetworkManager/system-connections/ file still exists after the connection is deleted ? Looks like we don't need the `self.remove_config(nmconn_file)` at all.

Not sure why there such a rule. Active voice and present tense[1] are always preferred and there are plenty of examples[2] where 'we' is used.

Yeah I've seen some left-overs and better to clean them up to not cause issues when we reboot I guess.

[1] https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/How-not-to-write-a-Git-commit-message#:~:text=Don't%20use%20passive%20voice,first%20word%20in%20subject%20line. 
[2] https://review.opendev.org/c/openstack/os-net-config/+/825224

karthiksundaravel:
Hi Rabi,
The guidelines are mentioned here [1].
Can you please share us the type of devices/connection for which the connection file is persistent after deleting it using `nmcli con del`.

[1]  https://github.com/os-net-config/os-net-config/blob/master/CONTRIBUTING.rst

rabi:
> Hi Rabi, The guidelines are mentioned here [1]. Can you please share us the type of devices/connection for which the connection file is persistent after deleting it using `nmcli con del`.
> 
> [1] https://github.com/os-net-config/os-net-config/blob/master/CONTRIBUTING.rst

@karthiksundaravel I did not see anything in those guidelines that suggests not to use active voice (i.e "imperative" means active voice). Also, I've seen some leftover files when doing the debug with Gil Rosenberg (I have no time to reproduce it now) and I think it's better to have a fallback to cleanup the files (if leftover). Feel free to change the PR to your liking and merge. We need this functionality to work in 18.0 FR2 and current testing in CI is failing.

karthiksundaravel (approved):
LGTM. Thanks.

PR #179, opened 2025-02-17
iprules: ignore defined ip rules

vcandapp:
If the IP rules are not defined by NetworkManager, os-net-config will be unable to remove them, resulting in a Verification Error. With this pull request, os-net-config record the rules set by os-net-config and will ignore other rules set on host

(cherry picked from commit 482b6300f0a443c8888ca43aeb59c030150828c0)


karthiksundaravel (approved):
LGTM. Thanks

PR #180, opened 2025-02-17
[DNM]migrate: Refactor the approach for migration of providers

karthiksundaravel:
The earlier approach uses the options like `enable_migration` and `clean_migration`. With the usage of reboot in the migration process, these options are no longer needed. The refactored approach is to backup and remove the network configurations, driver bindings, dpdk map files, sriov map files configured by the present provider and then create a service file which shall start with the new provider. After this the node is rebooted and when the system comes up the `os-net-config-migrate` service will configure with the provider. If the configuration with the new provider fails, then `--rollback` option allows restoring the configurations from the backed up config files.

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -73,6 +73,13 @@ def parse_opts(argv):
                         """configuration during the migration.""",
                         choices=_PROVIDERS.keys(),
                         default=None)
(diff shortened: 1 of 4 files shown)


karthiksundaravel:
PR# https://github.com/os-net-config/os-net-config/pull/209 is raised to handle the same

PR #182, opened 2025-02-17
sriov: skip PF configuration when its attached to guest

vcandapp:
In scenarios where the PF is attached to the guest, the PF device will not be visible in the hosts net subsystem. And if os-net-config is rerun in that scenario, the NIC mapping would fail. Checks shall be added so that the situation where the PF is attached to the guest is detected and handled.

(cherry picked from commit 91659a82a823c67cfb090613fbf4e13d574bcce2)


karthiksundaravel (approved):
LGTM. Thanks

PR #186, opened 2025-02-17
schema: route schema shall support nexthop and next_hop

vcandapp:
The current route schema requires `next_hop` when `ip_netmask` is configured and `nexthop` when `destination` is configured. But since the schema validation is ignored until recently and the object parser parses both `nexthop` and `next_hop` alike, the schema validation failure for the same did not fail os-net-config. Now the schema for `route` object is modified to support both `nexthop` and `next_hop` and hence the schema validation will not fail on existing deployments for the same.

(cherry picked from commit 1147937778cad911bddc941d939beafad4da6ce2)

### DIFF
--- a/os_net_config/schema.yaml
+++ b/os_net_config/schema.yaml
@@ -168,12 +168,15 @@ definitions:
 
     route:
         type: object
-        oneOf:
-        - properties:
+        properties:
             next_hop:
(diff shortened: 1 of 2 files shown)


karthiksundaravel (approved):
LGTM. Thanks

PR #188, opened 2025-02-26
bond: ignore order for dpdkbond

vcandapp:
When checking diff of DPDKBond data, the interfaces can be in any order. No of elements only matter

### DIFF
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -201,6 +201,12 @@ class IfcfgNetConfig(os_net_config.NetConfig):
         changed_values = {}
         for key in ifcfg_data_old:
             if key in ifcfg_data_new:
+                if key in ['BOND_IFACES', 'OVS_EXTRA']:
+                    if sorted(ifcfg_data_old[key].split()) == \
+                        sorted(ifcfg_data_new[key].split()):
+                        logger.debug("Ignoring order of list for %s", key)
+                        continue
+
                 if ifcfg_data_old[key].upper() != ifcfg_data_new[key].upper():
                     changed_values[key] = 'modified'
             else:
@@ -1560,7 +1566,10 @@ class IfcfgNetConfig(os_net_config.NetConfig):
                 ivs_uplinks.append(interface_name)
             if "NFVSWITCH_BRIDGE" in iface_data:
                 nfvswitch_interfaces.append(interface_name)
-            if utils.diff(interface_path, iface_data):
+            old = self.parse_ifcfg(common.get_file_data(interface_path))
+            new = self.parse_ifcfg(iface_data)
+            changes = self.enumerate_ifcfg_changes(old, new)
+            if len(changes.values()):
                 if self.ifcfg_requires_restart(interface_path, iface_data):
                     restart_interfaces.append(interface_name)
                     # Openvswitch needs to be restarted when OVSDPDKPort or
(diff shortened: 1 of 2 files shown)


os_net_config/impl_ifcfg.py
```diff
@@ -201,6 +201,12 @@ def enumerate_ifcfg_changes(self, ifcfg_data_old, ifcfg_data_new):
         changed_values = {}
         for key in ifcfg_data_old:
             if key in ifcfg_data_new:
+                if key in ['BOND_IFACES', 'OVS_EXTRA']:
```
karthiksundaravel, line 204: Can this be performed irrespective of the keys in ['BOND_IFACES', 'OVS_EXTRA']

os_net_config/impl_ifcfg.py
```diff
@@ -201,6 +201,12 @@ def enumerate_ifcfg_changes(self, ifcfg_data_old, ifcfg_data_new):
...
+                    if sorted(ifcfg_data_old[key].split()) == \
```
karthiksundaravel, line 205: Convert it to upper case for comparison

os_net_config/impl_ifcfg.py
```diff
@@ -201,6 +201,12 @@ def enumerate_ifcfg_changes(self, ifcfg_data_old, ifcfg_data_new):
         changed_values = {}
         for key in ifcfg_data_old:
```
karthiksundaravel, line 202: A similar logic shall be added in utils.diff() [1]  as well so that the first level comparison will ignore the changes.
[1]  https://github.com/os-net-config/os-net-config/blob/cfaceec3a9a02134816ed44ede53ff371a50f2cd/os_net_config/utils.py#L220

karthiksundaravel (changes requested):
I like the plan to compare the contents, but its better if we could do it in utils.diff() as well and the test cases shall use utils.diff() instead of calling enumerate_ifcfg_changes()

os_net_config/impl_ifcfg.py
vcandapp: Since ifcfg-* scripts are case-sensitive, can't ignore case here. Eg. ifcfg-DPDK0  and ifcfg-dpdk0 are valid:

/etc/sysconfig/network-scripts/ifcfg-DPDK0 
# This file is autogenerated by os-net-config
DEVICE=DPDK0
ONBOOT=yes
HOTPLUG=no
NM_CONTROLLED=no
PEERDNS=no
DEVICETYPE=ovs
TYPE=OVSDPDKPort
OVS_BRIDGE=BR-LINK0

and

/etc/sysconfig/network-scripts/ifcfg-dpdk0 
# This file is autogenerated by os-net-config
DEVICE=dpdk0
ONBOOT=yes
HOTPLUG=no
NM_CONTROLLED=no
PEERDNS=no
DEVICETYPE=ovs
TYPE=OVSDPDKPort
OVS_BRIDGE=br-link0

os_net_config/impl_ifcfg.py
vcandapp: Will limit the scope of this PR to this JIRA and update with a follow-up patch for other key elements

os_net_config/impl_ifcfg.py
dsneddon: If I understand correctly ifcfg files are case insensitive. These are the same functionality:
ONBOOT=YES
ONBOOT=yes

So if the only changes are case we don't want to consider it modified.

Please let me know if you think I'm incorrect here.

dsneddon (changes requested):
Please fix issues which are causing PEP8 to fail and address comments.

os_net_config/impl_ifcfg.py
vcandapp: Updated

dsneddon (approved):

PR #190, opened 2025-03-10
nmstate: use dispatcher scripts to configure unhandled ethtool opts

karthiksundaravel:
Certain ethtool opts like `-U`, `--set-priv-flags` are not yet supported by nmstate schema. Such ethtool opts are configured via NetworkManager-dispatcher scripts which directly calls the ethtool command when the device state changes to `up`.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -52,14 +52,17 @@ logger = logging.getLogger(__name__)
 # Import the raw NetConfig object so we can call its methods
 netconfig = os_net_config.NetConfig()
 
+DISPATCHER_SCRIPT_PREFIX = r"""
+set +e
+set -x
+"""
+
 # Script used to bind the VFs with appropriate drivers.
 # VFs specified in `dpdk_vfs` will be bound with vfio-pci.
 # VFs specified in `linux_vfs` will be bound with their default drivers
 _VF_BIND_DRV_SCRIPT = r'''
 dpdk_vfs="{dpdk_vfs}"
 linux_vfs="{linux_vfs}"
-set +e
-set -x
 
 for vfid in $dpdk_vfs $linux_vfs; do
     vf_pci_id=$(readlink "/sys/class/net/$1/device/virtfn$vfid") &&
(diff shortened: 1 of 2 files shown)


rjarry (dismissed):

dsneddon:
Recheck

os_net_config/impl_nmstate.py
```diff
@@ -146,6 +152,10 @@ def is_dict_subset(superset, subset):
     if superset == subset:
         return True
     if superset and subset:
+        if DISPATCH in superset.keys() and \
+            DISPATCH not in subset.keys():
+            return False
```
dsneddon, line 157: Not crazy about the indentation here. Let's see if it passes PEP8 tests.

os_net_config/impl_nmstate.py
```diff
@@ -1056,18 +1066,22 @@ def add_ethtool_config(self, iface_name, data, ethtool_options):
                 if command[1] not in accepted_dev_names:
                     msg = (
                         f"{iface_name}: Incorrect dev name found in "
-                        "{ethtool_opts}"
+                        f"{ethtool_opts}"
```
dsneddon, line 1069: Good catch

os_net_config/tests/test_impl_nmstate.py
```diff
@@ -309,6 +309,9 @@ def get_nmstate_ethtool_opts(self, name):
             self.provider.interface_data[name][Ethernet.CONFIG_SUBTREE]
         data[Ethtool.CONFIG_SUBTREE] = \
             self.provider.interface_data[name][Ethtool.CONFIG_SUBTREE]
+        if 'dispatch' in self.provider.interface_data[name].keys():
```
dsneddon, line 312: This could have been handled with a shorter key than "dispatch" like "dp" or "sh", but I support the use of the \ continuation for readability. 

Another thing you can do to fit things on one line is to assign the base object to a handle, like:

pvdr = self.provider

Then you can use:

pvdr.interface_data[...

dsneddon (approved):
Looks good, a few stylistic comments but if it passes tests I'm good with this as-is.

dsneddon (approved):

PR #191, opened 2025-03-24
nmstate: bind scripts fail when drivers_autoprobe is disabled

karthiksundaravel:
The bind scripts used in network manager dispatcher fails to handle the scenario when the VFs are not bound with any driver. The path /sys/bus/pci/devices/$vf_pci_id/driver will not be present when the sriov_drivers_autoprobe is disabled. The same is handled with this patch.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -65,7 +65,7 @@ dpdk_vfs="{dpdk_vfs}"
 linux_vfs="{linux_vfs}"
 
 for vfid in $dpdk_vfs $linux_vfs; do
-    vf_pci_id=$(readlink "/sys/class/net/$1/device/virtfn$vfid") &&
+    vf_pci_id=$(readlink -ve "/sys/class/net/$1/device/virtfn$vfid") &&
     vf_pci_id=$(basename "$vf_pci_id") &&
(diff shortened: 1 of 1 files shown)


rjarry (approved):

vcandapp (approved):
LGTM

os_net_config/impl_nmstate.py
```diff
@@ -75,8 +75,8 @@
     else
         driver="$def_driver"
     fi &&
-    cur_drv=$(readlink "/sys/bus/pci/devices/$vf_pci_id/driver") &&
-    cur_drv=$(basename "$cur_drv") &&
+    cur_drv=$(readlink "/sys/bus/pci/devices/$vf_pci_id/driver" 2>/dev/null) &&
+    cur_drv=$(basename "$cur_drv")
     if ! [ "$cur_drv" = "$driver" ]; then
```
dsneddon, line 80: Isn't the logic here changed with the removal of the && in L79, so now L80 is always interpreted and $cur_drv is always compared to $driver? But isn't it possible that $cur_driver has no value? This isn't a problem, is it?

dsneddon (commented):
I added a question to make sure that removing the && at the end of L79 doesn't potentially create broken edge cases

PR #192, opened 2025-03-26
sriov: wait for completion of driver binding

karthiksundaravel:
The path /sys/class/net/{pf_name}/device/virtfn{vfid}/net/{vf_name} will be available when the default VF driver is bound with the VF. When sriov_drivers_autoprobe=0, the drivers will not be bound with VF by default and hence the path will not be available. In NIC partitioning use cases, the driver will be bound by os-net-config. It is necessary to wait for the above path to be available in order to read the VF name. Few retries are made to check for the existence of the vf_path.

### DIFF
--- a/os_net_config/utils.py
+++ b/os_net_config/utils.py
@@ -565,19 +565,36 @@ def configure_sriov_vfs():
 
 
 def get_vf_devname(pf_name, vfid):
+    """Get the VF device name
+
+    :params pf_name: PF device name
+    :params vfid: VF id for which the name is required
+    :returns: VF name
+    """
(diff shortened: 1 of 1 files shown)


rjarry (changes requested):
5 seconds may not be enough in some edge cases. It may not hurt to wait more. How about 30 seconds before giving up?

rjarry (approved):

dsneddon:
recheck

vcandapp (approved):
LGTM

dsneddon (commented):
Thanks for this, I see the checks are passing now, good change.

PR #193, opened 2025-03-27
sriov: unit tests fails while accessing sriov_config.py

karthiksundaravel:
The unit tests fails while accessing the sriov_config.py from /var/lib/os-net-config/sriov_config.yaml. The paths are changed for the tests such that the map files could be created or read without root permissions.

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -183,6 +183,9 @@ def get_file_data(filename):
 
 
 def write_yaml_config(filepath, data):
+    if get_noop():
+        logger.info("Writing file %s with content %s", filepath, data)
+        return
     os.makedirs(os.path.dirname(filepath), exist_ok=True)
     with open(filepath, 'w') as f:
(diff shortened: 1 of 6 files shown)


dsneddon (dismissed):

os_net_config/tests/test_impl_nmstate.py
```diff
@@ -222,6 +222,10 @@ def setUp(self):
         super(TestNmstateNetConfig, self).setUp()
         common.set_noop(True)
 
+        rand = str(int(random.random() * 100000))
+        common.DPDK_MAPPING_FILE = '/tmp/dpdk_mapping_' + rand + '.yaml'
+        common.SRIOV_CONFIG_FILE = '/tmp/sriov_config_' + rand + '.yaml'
+
```
vcandapp, line 228: We might need to remove these stubs in the  tearDown() method

vcandapp (changes requested):

PR #195, opened 2025-04-01
sriov: sriov: unit tests fails while accessing sriov_config.py

karthiksundaravel:
The unit tests fails while accessing /var/lib/os-net-config/sriov_config.yaml. Also the temporary files corresponding to sriov_config.yaml, rep-link-name.sh 80-persistent-os-net-config.rules are removed during tearDown()


dsneddon (approved):
This looks good to me, pending successful RDO checks

rjarry (approved):

PR #197, opened 2025-04-04
nmstate: bind scripts fail when drivers_autoprobe is disabled

vcandapp:
The bind scripts used in network manager dispatcher fails to handle the scenario when the VFs are not bound with any driver. The path /sys/bus/pci/devices/$vf_pci_id/driver will not be present when the sriov_drivers_autoprobe is disabled. The same is handled with this patch.

(cherry picked from commit 64af7270ae14f5d00ff80d969b96f6f75f783267)


karthiksundaravel (approved):
lgtm

PR #198, opened 2025-04-05
cleanup: stop blindly removing nmconnection files

dsneddon:
Stops removing NetworkManager connections from the ifcfg provider. The logic in the NetworkManager cleanup is too indiscriminate and can cause removal of connections that don't need to be removed, should not be removed, and/or can cause spurious errors. A follow-up patch will
ensure there are no conflicts in a safe way that allows for peaceful co-existence of network-scripts and nmstate.

Suggested-by: Karthik Sundaravel <ksundara@redhat.com>
Fixes:  https://issues.redhat.com/browse/OSPRH-15443

### DIFF
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -19,7 +19,6 @@ import itertools
 import logging
 import netaddr
 import os
-from pathlib import Path
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


vcandapp (approved):
LGTM

PR #203, opened 2025-04-14
noop: fix noop run for nic partitioning use case

vcandapp:
The noop dry run attempts to write to the sriov map for SR-IOV or NIC partitioning deployments, which results in failure for non privileged users. The fix is to avoid writing to the map files during dry run.

(cherry picked from commit 7c0f68c8563843c808f90a379cde94340e5a58ff)


karthiksundaravel (approved):
lgtm !

PR #204, opened 2025-04-14
sriov: unit tests fails while accessing sriov_config.yaml

vcandapp:
The unit tests fails while accessing /var/lib/os-net-config/sriov_config.yaml. Also the temporary files corresponding to sriov_config.yaml, rep-link-name.sh 80-persistent-os-net-config.rules are removed during tearDown()

(cherry picked from commit 0c3447cee32a4708fde78ba71f0c65ad51472b6d)


karthiksundaravel (approved):
lgtm

karthiksundaravel (approved):

PR #206, opened 2025-04-23
sriov: intermittent exception seen in sriov_config service

vcandapp:
When os-net-config-sriov is run eg. during uprade/update, setting of PF config like autoprobe fails. Adding udev check to wait for the PF creation

Closes: https://issues.redhat.com/browse/OSPRH-15126

(cherry picked from commit 217a3f02ccd9e49b2c03bb03ca0ac8060360faa9)

### DIFF
--- a/os_net_config/sriov_config.py
+++ b/os_net_config/sriov_config.py
@@ -166,6 +166,28 @@ def _wait_for_vf_creation(pf_name, numvfs):
     logger.info("%s: Required VFs are created", pf_name)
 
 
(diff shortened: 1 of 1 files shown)


vcandapp:
> Looks good, thanks. We should maybe consider a log message for _wait_for_sysfs_creation() in both the trunk and stable/wallaby branches in a follow-up patch, but it may not be needed.

Follow-up patch: https://github.com/os-net-config/os-net-config/pull/208

karthiksundaravel (approved):

PR #207, opened 2025-04-24
sriov: driver binding for VF before enabling promisc

karthiksundaravel:
The required driver for the VF shall be bound before applying the promisc settings. This is required to handle  NIC partitioning while sriov_drivers_autoprobe is disabled.


rjarry (dismissed):
LGTM after rebasing on master

rjarry (approved):

PR #209, opened 2025-05-06
migrate: refactor the approach for migration of providers

karthiksundaravel:
The earlier approach uses the options like `enable_migration` and `clean_migration`. The refactored approach uses purge-provider cli option to remove the network configurations, driver bindings created using the provider. Backup of the configurations could be optionally made by the providers. After this, the migration could be proceeded further by re-running os-net-config with the new provider.


os_net_config/cli.py
```diff
@@ -340,15 +324,27 @@ def main(argv=sys.argv, main_logger=None):
             obj = objects.object_from_json(iface_json)
             purge_provider.del_object(obj)
         purge_provider.destroy()
```
abhiramnarayana, line 326: Do we need to take care or the case this destroy or del_object fails? Do we need to return 1 if so?

os_net_config/cli.py
```diff
@@ -56,21 +56,19 @@ def parse_opts(argv):
...
-                        help="""Enable a migration from one provider."""
-                        """Cleans the network configurations created by """
-                        """the specified provider and migrates the same """
-                        """network configuration to the desired provider."""
-                        """There shall be no change in the input network """
-                        """configuration during the migration.""",
+                        help="""Cleans the network configurations created """
+                        """by the specified provider. There shall be no """
+                        """change in the input network config.yaml during """
+                        """the purge of the provider""",
```
vcandapp, line 71: Will we assume any default for this ? Also not all supported proivder(s) can be purged. So, we can add """"One of: ifcfg, nmstate only""""

os_net_config/cli.py
```diff
@@ -336,19 +303,102 @@ def main(argv=sys.argv, main_logger=None):
         common.reset_dcb_map()
 
     if purge_provider:
-        for iface_json in iface_array:
+        ret_code = unconfig_provider(
```
vcandapp, line 306: If we assume default is "ifcfg", then we can do check like below for provider

os_net_config/cli.py
```diff
@@ -336,19 +303,102 @@ def main(argv=sys.argv, main_logger=None):
...
+        dcb_apply.apply()
+
+    if opts.detailed_exit_codes or ret_code == 1:
+        return ret_code
+    else:
+        return 0
+
+
+def unconfig_provider(provider_name,
+                      iface_array,
```
vcandapp, line 362: Is partial purge allowed ? If not, we might not need the iface_array - whatever configured in system is the list

karthiksundaravel: It is hard to back track the configuration applied by os-net-config. There is no backup of the config.yaml applied during the previous successful run. The expectation is to use the same config.yaml during purge_provider.

dsneddon (dismissed):
Please make sure the comments left by Viji are addressed otherwise approved.

os_net_config/cli.py
```diff
@@ -404,81 +485,48 @@ def main(argv=sys.argv, main_logger=None):
...
 
-        files_changed = provider.apply(cleanup=opts.cleanup,
-                                       activate=not opts.no_activate)
+        files_changed = provider.apply(cleanup=cleanup,
+                                       activate=not no_activate)
         logger.info(
-            "Succesfully applied the network configuration with "
-            "%s provider",
-            opts.provider,
+            "%s: Succesfully configured %s", provider_name, config_name
```
abhiramnarayana, line 508: Fix typo

os_net_config/cli.py
```diff
@@ -295,60 +262,174 @@ def main(argv=sys.argv, main_logger=None):
...
+                      root_dir,
+                      noop,
+                      ):
+    logger.info("%s: Performing unconfig", provider_name)
+    try:
+        purge_provider = load_provider(provider_name, noop,
+                                       root_dir)
+    except ImportError as e:
+        logger.error(
+            "cannot load purge provider %s: %s", purge_provider, e
```
abhiramnarayana, line 398: purge_provider will wont be there in exception case I think right?

os_net_config/cli.py
```diff
@@ -335,20 +301,136 @@ def main(argv=sys.argv, main_logger=None):
...
+            opts.purge_provider,
+            iface_array,
+            opts.root_dir,
+            opts.noop
+        )
+        if purge_ret != 0:
+            main_logger.error(
+                "Failed to purge %s provider", opts.purge_provider
+            )
+            return purge_ret
```
vcandapp, line 315: With more flags being added, some combinations might be mutually exclusive now. Eg. Is "os-net-config --cleanup --purge-provider ifcfg" allowed ? Can we identify such cases and fail ?

os_net_config/cli.py
```diff
@@ -208,38 +206,6 @@ def main(argv=sys.argv, main_logger=None):
...
-            opts.provider = "ifcfg"
-        elif os.path.exists('%s/etc/network/' % opts.root_dir):
-            opts.provider = "eni"
-        else:
-            main_logger.error("Unable to set provider for this operating "
-                              "system.")
-            return 1
-
-    if opts.purge_provider:
-        if opts.purge_provider == opts.provider:
```
vcandapp, line 232: This check seems to be missed - this might be needed

karthiksundaravel: It is intentionally omitted. It allows me to clean the current provider and reconfigure with the same provider :) @dsneddon  @vcandapp  Your thoughts.

vcandapp (approved):
LGTM

PR #210, opened 2025-05-06
migrate: add utils to destroy dpdk ports

karthiksundaravel:
Add utility methods to detach the dpdk ports from ovs and unset the
driver overrides configured for the dpdk ports.

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -508,6 +508,22 @@ def wait_for_vf_driver_binding(pf, vfs, req_driver):
                 )
 
 
+def unset_driverctl_override(pci_address):
+    cmd = ["driverctl", "unset-override", pci_address]
+    try:
+        logger.info("%s: running %s", pci_address, " ".join(cmd))
+        out, err = processutils.execute(*cmd)
+        if err:
+            logger.error("%s: Failed to unbind dpdk interface.", pci_address)
+            return 1
+        return 0
+    except processutils.ProcessExecutionError as exc:
+        logger.error(
+            "%s: Failed to bind interface with dpdk. %s", pci_address, exc
+        )
+        return 1
+
+
 def set_driverctl_override(pci_address, driver):
     if driver is None:
         logger.info("%s: Driver override is not required.", pci_address)
--- a/os_net_config/utils.py
+++ b/os_net_config/utils.py
@@ -236,6 +236,18 @@ def update_dpdk_map(ifname, driver):
         _update_dpdk_map(ifname, pci_address, mac_address, driver)
 
 
+def unbind_dpdk_interfaces(pci_address):
+    iface_driver = common.get_pci_device_driver(pci_address)
+    if iface_driver == "vfio-pci":
+        return common.unset_driverctl_override(pci_address)
+    else:
+        logger.info(
+            "%s: not bound with vfio-pci and hence not removing the override",
+            pci_address,
+        )
+        return 0
+
+
 def bind_dpdk_interfaces(ifname, driver, noop):
     if common.is_mellanox_interface(ifname) and 'vfio-pci' in driver:
         msg = ("For Mellanox NIC %s, the default driver vfio-pci "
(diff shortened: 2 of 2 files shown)


os_net_config/utils.py
```diff
@@ -236,6 +236,17 @@ def update_dpdk_map(ifname, driver):
         _update_dpdk_map(ifname, pci_address, mac_address, driver)
 
 
+def unbind_dpdk_interfaces(pci_address):
+    iface_driver = common.get_pci_device_driver(pci_address)
+    if iface_driver == "vfio-pci":
+        common.unset_driverctl_override(pci_address)
```
abhiramnarayana, line 242: We need to propogate the return value from the unset_driverctl_overrifde right. should we be doing
return common.unset_driverctl_override(pci_address)  ?

os_net_config/utils.py
```diff
@@ -286,6 +297,33 @@ def bind_dpdk_interfaces(ifname, driver, noop):
...
+
+
+def detach_dpdk_interfaces(pci_address):
+    cmd = ["ovs-appctl", "netdev-dpdk/detach", pci_address]
+    try:
+        logger.info("%s: running %s", pci_address, " ".join(cmd))
+        out, err = processutils.execute(*cmd)
+    except processutils.ProcessExecutionError as exc:
+        logger.error("%s: Failed to detach. Err: %s", pci_address, exc)
+        return -1
```
abhiramnarayana, line 324: should this be return 1 (similar to function unset_driverctl_override where we return 1 in case of err)

os_net_config/utils.py
```diff
@@ -286,6 +298,40 @@ def bind_dpdk_interfaces(ifname, driver, noop):
...
+    dpdk_map = common.get_dpdk_map()
+    for dpdk_nic in dpdk_map:
+        if dpdk_nic["name"] == iface:
+            err = detach_dpdk_interfaces(dpdk_nic["pci_address"])
+            if err:
+                logger.warning(
+                    "%s: Failed to detach dpdk interface",
+                    dpdk_nic["name"],
+                )
+            err = unbind_dpdk_interfaces(dpdk_nic["pci_address"])
```
vcandapp, line 311: Shud we safely break here also for error ?

os_net_config/common.py
```diff
@@ -508,6 +508,22 @@ def wait_for_vf_driver_binding(pf, vfs, req_driver):
...
+        out, err = processutils.execute(*cmd)
+        if err:
+            logger.error("%s: Failed to unbind dpdk interface.", pci_address)
+            return 1
+        return 0
+    except processutils.ProcessExecutionError as exc:
+        logger.error(
+            "%s: Failed to bind interface with dpdk. %s", pci_address, exc
+        )
+        return 1
```
vcandapp, line 524: can this be "raise" instead of return ?

vcandapp (changes requested):

os_net_config/utils.py
karthiksundaravel: The detach can fail if the pci address is no longer attached to the port, which means it is already detached.  If it is already detached, then we are good to proceed.

os_net_config/common.py
karthiksundaravel: This would allow the execution to continue and not be stopped due to single failure. This will be part of cleanup routine so it is best that the cleanup is not halted for a single failure.

os_net_config/utils.py
```diff
@@ -236,6 +236,18 @@ def update_dpdk_map(ifname, driver):
         _update_dpdk_map(ifname, pci_address, mac_address, driver)
 
 
+def unbind_dpdk_interfaces(pci_address):
+    iface_driver = common.get_pci_device_driver(pci_address)
+    if iface_driver == "vfio-pci":
```
vcandapp, line 241: What about nic-partitioned "mlx5_core" driver ?

dsneddon (approved):
This looks very good. Thanks!

PR #211, opened 2025-05-06
migrate: backup the dpdk,sriov map files

karthiksundaravel:
During migration from ifcfg to nmstate provider the config files sriov_config.yaml, dpdk_mapping.yaml, 70-os-net-config-sriov.rules shall be backed up.

### DIFF
--- a/os_net_config/utils.py
+++ b/os_net_config/utils.py
@@ -18,6 +18,7 @@ import glob
 import logging
 import os
 import re
+import shutil
 import time
 import yaml
 
@@ -647,6 +648,22 @@ def get_vf_devname(pf_name, vfid):
     # The VF's actual device name shall be the only directory seen in the path
     # /sys/class/net/<pf_name>/device/virtfn<vfid>/net
     return vf_nic[0]
+
+
+def backup_map_files(backup_path):
+    logger.debug("Backing up DPDK, SR-IOV, and mapping files.")
+    src_files = [common.DPDK_MAPPING_FILE,
+                 common.SRIOV_CONFIG_FILE,
+                 sriov_config._UDEV_LEGACY_RULE_FILE
+                 ]
+    for src in src_files:
+        if os.path.exists(src):
+            src_name = os.path.basename(src)
+            bkup_file = os.path.join(backup_path, src_name)
+            logger.info("Copying %s -> %s", src, bkup_file)
+            shutil.copy(src, bkup_file)
+        else:
+            logger.warning("%s does not exist", src)
 
 
 def restart_vpp(vpp_interfaces):


os_net_config/utils.py
```diff
@@ -649,6 +650,21 @@ def get_vf_devname(pf_name, vfid):
     return vf_nic[0]
 
 
+def backup_map_files(backup_path):
+    src_files = [common.DPDK_MAPPING_FILE,
```
dsneddon, line 654: Maybe add a debug message saying "Backing up DPDK, SR-IOV, and mapping files." here.

os_net_config/utils.py
```diff
@@ -649,6 +650,21 @@ def get_vf_devname(pf_name, vfid):
...
+                 sriov_config._UDEV_LEGACY_RULE_FILE
+                 ]
+    for src in src_files:
+        if os.path.exists(src):
+            src_name = os.path.basename(common.DPDK_MAPPING_FILE)
+            bkup_file = os.path.join(backup_path, src_name)
+            logger.info("Moving %s -> %s", src, bkup_file)
+            shutil.move(src, bkup_file)
+        else:
+            logger.warning("%s is absent", src)
```
dsneddon, line 665: I would say "%s does not exist." rather than "%s is absent."

os_net_config/utils.py
```diff
@@ -649,6 +650,21 @@ def get_vf_devname(pf_name, vfid):
...
 
+def backup_map_files(backup_path):
+    src_files = [common.DPDK_MAPPING_FILE,
+                 common.SRIOV_CONFIG_FILE,
+                 sriov_config._UDEV_LEGACY_RULE_FILE
+                 ]
+    for src in src_files:
+        if os.path.exists(src):
+            src_name = os.path.basename(common.DPDK_MAPPING_FILE)
+            bkup_file = os.path.join(backup_path, src_name)
```
dsneddon, line 661: Can we back this up to a time stamped name rather than a fixed name just in case the operation had to be run more than once? It would be a shame to end up overwriting an only version of the file if migration failed the first time and had to be run again. Or is this scenario unlikely to occur?
I haven't gone back to check if the calling function already puts a timestamp in the backup path, in which case ignore this comment.

dsneddon (dismissed):
Approved only if my advice about better logging messages does not apply.

os_net_config/utils.py
karthiksundaravel: The caller is yet to be added. Its there in PR 212. I think we can create a directory with timestamp and move the files  there. If that is fine, this function can stand as it is.

dsneddon:
recheck

dsneddon (approved):
I'm satisfied that my concerns have been addressed. LGTM

os_net_config/utils.py
```diff
@@ -649,6 +650,22 @@ def get_vf_devname(pf_name, vfid):
     return vf_nic[0]
 
 
+def backup_map_files(backup_path):
+    logger.debug("Backing up DPDK, SR-IOV, and mapping files.")
+    src_files = [common.DPDK_MAPPING_FILE,
+                 common.SRIOV_CONFIG_FILE,
+                 sriov_config._UDEV_LEGACY_RULE_FILE
```
vcandapp, line 657: There will be on DCB config file - but ignore if it will be covered in separate PR. DCB is only TP as of now

karthiksundaravel: DCB is common for both the providers so it remains the same irrespective of migration

karthiksundaravel:
Merging it since @dsneddon gave his prior approval.

PR #212, opened 2025-05-08
migrate: purge ifcfg provider

karthiksundaravel:
Adding support for purging the network configurations created by ifcfg provider. The network interfaces configured by the ifcfg provider would be brought down. The DPDK ports will be detached from ovs bridge and driver override will be removed. The sriov_numvfs of SR-IOV interfaces will be reset. All the ifcfg files created by the ifcfg provider would be backed up and removed from the /etc/sysconfig/network-scripts path


os_net_config/impl_ifcfg.py
```diff
@@ -1048,7 +1058,7 @@ def del_ovs_user_bridge(self, bridge):
         :param bridge: The OvsUserBridge object to be deleted.
         """
         logger.info("%s: Deleting ovs user bridge", bridge.name)
-        self._del_common(bridge)
+        self.del_device["ovs_bridge"].append(bridge.name)
```
vcandapp: Can del_bridge() and del_user_bridge() be optimised as single method ?

os_net_config/impl_ifcfg.py
```diff
@@ -204,8 +204,18 @@ def __init__(self, noop=False, root_dir=''):
...
         self.renamed_interfaces = {}
         self.bond_primary_ifaces = {}
-        self.remove_iface = []
-        self.remove_sriov_pfs = []
+        self.del_device = {
+            "iface": [],
+            "vlan": [],
+            "ovs_bond": [],
+            "linux_bond": [],
+            "linux_bridge": [],
```
vcandapp: Do we support config/del of linux_bridge ?

karthiksundaravel: using linux_bridge is not recommended for RHOSO, but os-net-config could still support linux_bridge

karthiksundaravel:
recheck

abhiramnarayana (approved):
LGTM

PR #215, opened 2025-05-09
fallback: add support for falling back to a default configuration

karthiksundaravel:
fallback: add support for falling back to a default configuration

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -209,11 +209,8 @@ def main(argv=sys.argv, main_logger=None):
     common.logger_level(main_logger, opts.verbose, opts.debug)
     main_logger.info("Using config file at: %s", opts.config_file)
     iface_array = []
-    configure_sriov = False
-    sriovpf_bond_ovs_ports = []
     provider = None
     purge_provider = None
-    files_changed = {}
 
     if opts.purge_provider:
         try:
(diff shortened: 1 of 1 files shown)


dsneddon (changes requested):
Please add a sample fallback config alongside the JSON/YAML samples in the sample directory.

os_net_config/cli.py
```diff
@@ -367,8 +381,16 @@ def main(argv=sys.argv, main_logger=None):
             "%s: Failed to configure network_config. ",
             opts.provider,
         )
-        return get_exit_code(opts.detailed_exit_codes,
-                             onc_ret_code | ExitCode.NETWORK_CONFIG_FAILED)
+        onc_ret_code |= ExitCode.NETWORK_CONFIG_FAILED
```
vcandapp, line 384: There might be two scenarios here - fallback config present / absent. Should we handle both ? Eg. in present case - can we reset the error code ?

os_net_config/cli.py
```diff
@@ -367,8 +381,16 @@ def main(argv=sys.argv, main_logger=None):
...
+        ret_code = safe_fallback(
+            opts.provider,
+            fb_config,
```
vcandapp, line 387: Add safe check for fb_config = []

os_net_config/cli.py
```diff
@@ -699,5 +721,32 @@ def get_iface_config(
...
+        root_dir,
+        noop,
+        no_activate,
+        False,
+    )
+    if ret == ExitCode.ERROR:
+        logger.error("%s: failed to configure fallback_config", provider)
+        return ExitCode.FALLBACK_FAILED
+    else:
+        logger.info("%s: fallback_config is completed", provider)
```
vcandapp, line 747: Will an extra debug log "network_config failed and it needs to be corrected..." ?

os_net_config/cli.py
```diff
@@ -699,5 +721,32 @@ def get_iface_config(
     return iface_array
 
 
+def safe_fallback(provider,
+                  fb_config,
+                  no_activate,
+                  root_dir,
+                  noop):
+    if not fb_config:
```
vcandapp, line 729: Is it correct to assume the length of fb_config should be 1 ?

vcandapp (approved):
Approving the PR

PR #216, opened 2025-05-13
cli: add new flag to support clean_all nm conn

vcandapp:
When os-net-config-sriov is run after RHOSO node-prov with VLAN interface using cloud-init, there will be nm connections.
This needs to be cleaned, if ifcfg provider is used

Closes: https://issues.redhat.com/browse/OSPRH-16526


vcandapp:
recheck

vcandapp:
recheck /trigger-ci-gating

vcandapp:
Checking /trigger-ci-gating

PR #217, opened 2025-05-21
nmstate: cleanup the NM connections

karthiksundaravel:
The cleanup actions cannot be allowed on ovs interfaces, bridges and SR-IOV devices as removing these network interfaces will cause significant downtime to the networks and it some of these interfaces cannot be removed either. If there is a real need to remove these interfaces it would be better to re-configure the system again.


karthiksundaravel:
recheck

dsneddon:
Thank you very much for the contribution. It appears that this change would implement the changes required in this Jira issue:
https://issues.redhat.com/browse/OSPRH-15443?jql=assignee%20%3D%20currentUser()%20AND%20resolution%20%3D%20Unresolved%20ORDER%20BY%20summary%20ASC%2C%20updated%20DESC

os_net_config/common.py
```diff
@@ -273,6 +273,11 @@ def get_sriov_pfs():
     return [pf for pf in sriov_map if pf["device_type"] == "pf"]
 
 
+def get_sriov_pf_names():
```
vcandapp, line 276: Is is possible to re-use the existing " get_sriov_pfs()" and extract pf['name'] from it ?

vcandapp (changes requested):

os_net_config/impl_nmstate.py
```diff
@@ -721,21 +721,49 @@ def cleanup_all_ifaces(self, exclude_nics=[]):
...
-                state = {Interface.KEY: [iface]}
-                self.__dump_key_config(
-                    iface, msg=f"{iface[Interface.NAME]}: Cleaning up"
+            if iface.get(Interface.NAME) in exclude_nics:
+                logger.info(
+                    "%s: Interface name is excluded from cleanup",
+                    iface[Interface.NAME]
+                )
+                continue
+            if common.is_vf_by_name(iface.get(Interface.NAME)):
```
dsneddon, line 737: If we were only targeting Python 3.10 and up a match statement would be good here. I would have used if/elif statements, but this works too, it's just harder to read in my opinion.

dsneddon (approved):
A couple comments but overall I think this could merge as-is.

PR #222, opened 2025-06-06
ifcfg: add sysctl flags for ipv6 defroute

dsneddon:
This change adds IPv6-related flags to the ifcfg
configuration which define the IPv6 gateway explicitly
and cause IPv6 route advertisements to be ignored when
a static IPv6 default gateway is provided.


dsneddon:
recheck

os_net_config/impl_ifcfg.py
```diff
@@ -1687,6 +1700,8 @@ def apply(self, cleanup=False, activate=True, config_rules_dns=True):
             all_file_names.append(br_route_path)
             all_file_names.append(br_route6_path)
             all_file_names.append(br_rule_path)
+            if "IPV6_FORCE_ACCEPT_RA" in bridge_data:
```
vcandapp, line 1703: Is the check needed in above nfvswitch data ?

PR #226, opened 2025-06-13
nmstate: handle the removal of ethtool --set-priv-flags

karthiksundaravel:
The NetworkManager-dispatcher scripts are used for handling ethtool --set-priv-flags. When the ethtool --set-priv-flags are removed the dispatcher scripts shall be updated to reflect the change. A reboot of the machine is required to apply the change however.

(cherry picked from commit 4aa5b031ee36e3066c4f2a278209305c1fda02b8)


vcandapp (approved):
Cherry pick LGTM

PR #227, opened 2025-06-13
nmstate: avoid cleanup of ovs, sriov and other config.yaml interfaces

karthiksundaravel:
The cleanup actions cannot be allowed on ovs interfaces, ovs bridges, SR-IOV devices and other interfaces defined in config.yaml. If there is a real need to remove these interfaces it would be recommended to re-configure the system again instead of cleanup.

Closes: https://issues.redhat.com/browse/OSPRH-17034

(cherry picked from commit 805ec05f8553d5d3b19c839b507b4c0779c0d9cf)

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -367,7 +367,8 @@ def main(argv=sys.argv, main_logger=None):
         # os-net-config skips the ifup <ifcfg-pfs>, since the ifcfgs for PFs
         # wouldn't have changed.
         if configure_sriov:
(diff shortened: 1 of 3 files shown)


karthiksundaravel:
> I'm confused by this, wouldn't --cleanup cause any interfaces that are not defined in the config to have their configuration removed, regardless of what type of interface it is. Why do we need special handling of PF, etc.?

Dan, this will make nmstate at par with ifcfg. The cleanup of DPDKs and SR-IOVs are not handled for ifcfg or nmstate yet. We could target a new PR to handle the same after supporting migration related PRs.

PR #230, opened 2025-06-19
cleanup flag for nmstate interface removal

tyiying:
--cleanup flag is used to remove the interfaces that's not defined in os-net-config's config.yaml

However, in nmstate, we mistakenly deleted the interfaces that's not even configured by os-net-config, i.e. interfaces that was not in previous config.yaml got deleted. 

This PR is to restrict --cleanup to remove only interfaces that were configured by os-net-config. 

1. save nmstate current network config into '/var/lib/os-net-config/nmstate_files/network_config.yaml' once configuration is successful
2. keep using --cleanup flag, change cleanup_all_ifaces() function used by cleanup:
- remove calling of netinfo.show_running_config()[Interface.KEY],
- replace with function self.get_current_ifaces() to read in the current configured interfaces from '/var/lib/os-net-config/nmstate_files/network_config.yaml'


karthiksundaravel (changes requested):
`--cleanup` has a different use case for RHOSO where all the interfaces needs to be cleaned. We need to use a different argument for the use case described. Also I think we can make it independent of provider

yiytan:
Using remove_config worked too. But it need user to run external tools and update remove_config to achieve the selective cleanup. And this is only needed by nmstate, since in ifcfg, cleanup flag is doing the job. It compare all interface files being created by the new network_config against the files in /etc/sysconfig/network-scripts/ifcfg- which was created by os-net-config.

PR #231, opened 2025-06-20
nmstate: fix device type for lo

karthiksundaravel:
The device type for `lo` shall be loopback and not ethernet for it to be properly configured by nmstate.


os_net_config/impl_nmstate.py
```diff
@@ -106,6 +106,7 @@
 POST_ACTIVATION = 'post-activation'
 POST_DEACTIVATION = 'post-deactivation'
 DISPATCH = 'dispatch'
+LOOPBACK = "lo"
```
vcandapp, line 109: @karthiksundaravel , will there be only one lo interface in all our use cases ? Any chance we might have more lo0, lo1, etc ?

vcandapp (approved):
Thanks for the clarification. LGTM

PR #233, opened 2025-06-21
nmstate: fix device type for lo

karthiksundaravel:
The device type for `lo` shall be loopback and not ethernet for it to be properly configured by nmstate.

(cherry picked from commit 331803ee12356e0fde5cea031ec9b37adb96d75d)


vcandapp (approved):
This cherry-pick LGTM

PR #236, opened 2025-06-23
ifcfg: set ipv6 sysctls for static routing

dsneddon:
This change adds IPv6-related flags to the ifcfg configuration
 which define the IPv6 gateway explicitly and cause IPv6 route
advertisements to be ignored when a static IPv6 default
gateway is provided. Also, some minor cleanup and typo 
corrections are included.


os_net_config/utils.py
```diff
@@ -927,3 +927,12 @@ def ethtool_path():
         logger.warning("Could not execute /sbin/ethtool or /usr/sbin/ethtool")
         return False
     return ethtoolcmd
+
+
+def set_accept_ra_sysctl(iface):
+    """Set /proc/sys/net/ipv6/conf/<iface>/accept_ra """
+    filename = "/proc/sys/net/ipv6/conf/{}/accept_ra".format(iface)
+    if os.path.exists(filename):
+        logger.info("DEFROUTE=no: Set sysctl to ignore router advertisements")
```
abhiramnarayana, line 936: @dsneddon Do you want to include interface name in the log message for better traceability?

os_net_config/impl_ifcfg.py
```diff
@@ -275,7 +275,9 @@ def ifcfg_requires_restart(self, filename, new_data):
         permitted_changes = [
             'IPADDR', 'NETMASK',
             'MTU', 'ONBOOT', 'ETHTOOL_OPTS',
-            'DOMAIN', 'DNS1', 'DNS2'
+            'DOMAIN', 'DNS1', 'DNS2',
+            'IPV6_SET_SYSCTLS, IPV6_DEFAULTGW,',
+            'IPV6_DEFAULTDEV", "IPV6_FORCE_ACCEPT_RA'
```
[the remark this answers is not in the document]
dsneddon: Yes, fixed in next revision

os_net_config/utils.py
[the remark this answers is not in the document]
dsneddon: Fixing in next revision

os_net_config/utils.py
```diff
@@ -927,3 +927,12 @@ def ethtool_path():
...
+        with open(filename, 'w') as f:
+            f.write("0")
```
karthiksundaravel: is persistence across reboots needed ? What if there is a need to change this value to 1 later ?

karthiksundaravel (changes requested):

os_net_config/utils.py
dsneddon: I added a comment. This sets the sysctl immediately if the interface isn't restarted, but after that the ifcfg settings will cause the ifup script to set the sysctls.

karthiksundaravel (dismissed):
Thanks. LGTM.

PR #238, opened 2025-06-24
ovs: add ovs_extra option for interface and sriov_vf

karthiksundaravel:
The ovs_extra settings for ovs_port is required to configure `external-ids`. The schema for `interface` and `sriov_vf` types are modified to allow the inclusion of `ovs_extra` fields when these types are members of ovs_bridges.


os_net_config/impl_nmstate.py
```diff
@@ -1833,6 +1840,34 @@ def parse_ovs_extra(self, ovs_extras, name, data):
             for cmd_map in cfg_val_pair:
                 self._ovs_extra_cfg_val(ovs_extra_cmd, cmd_map, data)
 
+    def parse_ovs_extra_for_iface(self, ovs_extras, iface_name, data):
+        """Parse ovs extra for interface
+
+        :param ovs_extras: given ovs extra as string
```
dsneddon, line 1846: Suggestion: I know we have the same docstring other places, but it should probably be clear that ovs_extras is a list and not a string.

os_net_config/impl_nmstate.py
```diff
@@ -1833,6 +1840,34 @@ def parse_ovs_extra(self, ovs_extras, name, data):
...
+                      'sub_tree': [OvsDB.KEY, OvsDB.EXTERNAL_IDS],
+                      'nm_config': None,
+                      'nm_config_regex': r'^external-ids:(.+?)=',
+                      'value_pattern': r'^external-ids:.*=(.+?)$'},
+                     {'config': r'^external_ids:[\w+]',
+                      'sub_tree': [OvsDB.KEY, OvsDB.EXTERNAL_IDS],
+                      'nm_config': None,
+                      'nm_config_regex': r'^external_ids:(.+?)=',
+                      'value_pattern': r'^external_ids:.*=(.+?)$'}
+                     ]
```
dsneddon, line 1861: I'm not sure I understand why this iface_cfg is a list of these two identical dicts. Maybe add a docstring if you do another pass.

dsneddon (approved):
Only very minor docstring suggestions. Looks good otherwise.

os_net_config/impl_ifcfg.py
```diff
@@ -528,6 +528,11 @@ def _add_common(self, base_opt):
         elif isinstance(base_opt, objects.Interface):
             if base_opt.linkdelay:
                 data += "LINKDELAY=%s\n" % base_opt.linkdelay
+            if base_opt.bridge_name:
+                ovs_extra.extend(base_opt.ovs_extra)
```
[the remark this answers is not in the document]
karthiksundaravel: Yes, bond needs to be handled for nmstate scenario while its already supported in ifcfg.  PR:239 is raised for the same as a follow up patch.

dsneddon (approved):
This looks great. My only note is that I wonder if we need to test some more OVS extra commands, such as OVS-DPDK dpdk-lsc-interrupt or MTU settings. I know there are NFV clouds that use those settings. Perhaps we can make sure they are tested with nmstate in CI or do a follow up.

karthiksundaravel:
> This looks great. My only note is that I wonder if we need to test some more OVS extra commands, such as OVS-DPDK dpdk-lsc-interrupt or MTU settings. I know there are NFV clouds that use those settings. Perhaps we can make sure they are tested with nmstate in CI or do a follow up.

changing MTU is supported today, while `ovs-vsctl set interface <iface_name> options:dpdk-lsc-interrupt=true` is not supported yet in nmstate. We have to raise the change request with nmstate eventually.

PR #239, opened 2025-06-24
nmstate: handle ovs_extra for ovs bonds

karthiksundaravel:
Adding support to handle the ovs_extra fields for ovs bonds, dpdk bonds


vcandapp (approved):
LGTM - patch_port use case can be re-visited later with a follow-up patch

vcandapp (approved):
lgtm

PR #246, opened 2025-07-09
schema: schema support for remove_config

karthiksundaravel:
Adding the schema support to handle remove_config. The `remove_device` schema shall be mutually exclusive with the other interface types and can not be used inside `network_config`.

### DIFF
--- a/os_net_config/schema.yaml
+++ b/os_net_config/schema.yaml
@@ -1,5 +1,5 @@
 ---
-$schema: http://json-schema.org/draft-04/schema
+$schema: "http://json-schema.org/draft-07/schema"
 
 definitions:
     # base types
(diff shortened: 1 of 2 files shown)


os_net_config/schema.yaml
```diff
@@ -146,6 +146,29 @@ definitions:
         items:
             $ref: "#/definitions/address"
         minItems: 1
+    device:
+        type: object
+        properties:
```
vcandapp, line 151: Do we plan to allow additionalProperties ?

os_net_config/schema.yaml
```diff
@@ -1674,4 +1698,65 @@ items:
...
+        - contains:
+            $ref: "#/definitions/contrail_vrouter_dpdk"
+        - contains:
+            $ref: "#/definitions/linux_tap"
+        - contains:
+            $ref: "#/definitions/dcb_config"
+    then:
+      not:
+        contains:
+            $ref: "#/definitions/device"
```
vcandapp, line 1762: Do we need schema for parsing "remove_config:" ?

os_net_config/schema.yaml
karthiksundaravel: Not required. I'll force it to false.

PR #251, opened 2025-07-21
utils: add SR-IOV cleanup functions for PF and VF removal

karthiksundaravel:
- Add remove_sriov_entries_for_pf() to remove PF/VF entries from sriov map
- Add del_udev_rule_for_legacy_sriov_pf() to clean up legacy udev rules
- Add disable_sriov_config_service() to disable systemd service when needed
- Update reset_sriov_pf() to target specific PF instead of all PFs
- Auto-remove config files when maps become empty


os_net_config/utils.py
```diff
@@ -415,6 +415,29 @@ def get_totalvfs(iface_name):
...
+        if (item["device_type"] == "vf" and
+                item["device"].get("name") == pfname):
+            continue  # Skip this VF
+        if (item["device_type"] == "pf" and
+                item.get("name") == pfname):
+            continue  # Skip this PF
+        new_map.append(item)
+    if not new_map:
+        if os.path.exists(common.SRIOV_CONFIG_FILE):
+            os.remove(common.SRIOV_CONFIG_FILE)
```
abhiramnarayana, line 435: Should we add a log here?

os_net_config/utils.py
```diff
@@ -415,6 +415,29 @@ def get_totalvfs(iface_name):
     return -1
 
 
+def remove_sriov_entries_for_pf(pfname):
+    """Remove all VF and PF entries for the PF device from the SR-IOV map.
+
+    If the resulting map is empty, remove the SRIOV_CONFIG_FILE.
+    """
+    sriov_map = common.get_sriov_map()
```
abhiramnarayana, line 423: should we be calling with the pfname here ? common.get_sriov_map(pf_name)

os_net_config/sriov_config.py
```diff
@@ -256,33 +256,28 @@ def get_numvfs(ifname):
     return curr_numvfs
 
 
-def reset_sriov_pfs():
+def reset_sriov_pf(pf_name):
     """Reset the given PF
 
     Reset the numvfs for all the PFs configured.
```
vcandapp, line 262: This description needs update

os_net_config/sriov_config.py
```diff
@@ -256,33 +256,28 @@ def get_numvfs(ifname):
...
                 )
-
-
-def wipe_sriov_udev_files():
-    if os.path.exists(_UDEV_LEGACY_RULE_FILE):
-        logger.debug("Removing %s", _UDEV_LEGACY_RULE_FILE)
-        os.remove(_UDEV_LEGACY_RULE_FILE)
-        reload_udev_rules()
+            break
+    del_udev_rule_for_legacy_sriov_pf(pf_name)
```
vcandapp, line 280: We delete irrespective of exception here ?

os_net_config/sriov_config.py
```diff
@@ -595,6 +590,32 @@ def add_udev_rule_for_legacy_sriov_pf(pf_name, numvfs):
...
+        return
+    file_data = common.get_file_data(udev_file)
+    udev_lines = file_data.splitlines()
+    new_lines = [line for line in udev_lines if pattern not in line]
+    # Ignore comment lines and check if the file is empty
+    non_comment_lines = [
+        line for line in new_lines
+        if line.strip() and not line.strip().startswith("#")
+    ]
+    if not non_comment_lines:
```
vcandapp, line 610: Can we add some debug logs to track add/del of rules ?

os_net_config/sriov_config.py
karthiksundaravel: Will do

os_net_config/sriov_config.py
karthiksundaravel: Yes, if we remove the entry from the map, a reboot will prevent any reconfiguration. A reboot could be done during the migration

os_net_config/sriov_config.py
karthiksundaravel: Will do

PR #253, opened 2025-07-22
[DNM]ci: use pending status for ci-gate label instead of failing

karthiksundaravel:
Replace core.setFailed() with GitHub Status API to show pending status when CI is in-progress, providing clearer state communication.


karthiksundaravel:
Not required any further

PR #261, opened 2025-07-31
fix ipv6 accept_ra settings

dsneddon:
The IPV6_FORCE_ACCEPT_RA setting was being set improperly and this was causing the sysctl settings to not be set properly when writing ifcfg files for the ifcfg provider. This fix now sets IPV6_FORCE_ACCEPT_RA=no whenever there is a static IPv6 address on an interface. It is expected that routes will be set statically.

Note that this patch assumes that whenever a static IPv6 address is set that static routes will also be used and that all router advertisements will be ignored. This means that static IPv6 addresses may not be used on the same interface with EUI64 addresses, but os-net-config already assumes that DHCPv6, SLAAC, and static IPv6 addresses are mutually exclusive on a given interface. Although IPv6 supports using multiple address types on an interface this is generally not desired for server usage and for security reasons it is important to use static addresses and routes in cloud environments (to prevent traffic hijacking).

There are other relevant settings for IPv6 that may be used in ifcfg files, but they are not supported by the network init-scripts, only by NetworkManager when the NM_CONTROLLED parameter is set to true. A follow-up patch will add support for these fine-tuning adjustments with NetworkManager.


karthiksundaravel (approved):
LGTM ! Thanks.

PR #263, opened 2025-08-04
nmstate: implement error handling for ovs_extra bridge configurations

abhiramnarayana:
This patch implements comprehensive error detection and validation for ovs_extra bridge configurations, providing clear error messages for unsupported commands while maintaining full backward compatibility. This is patch 4 of a multi-patch series focused on OVS bridge ovs_extra enhancements.

This completes the bridge-focused portion of the ovs_extra enhancement series. Future patches will address DPDK and bond related ovs_extra configurations.

Assisted-by: cursor/claude-4-sonnet

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -1701,6 +1701,7 @@ class NmstateNetConfig(os_net_config.NetConfig):
         :raises ConfigurationError: Invalid ovs_extra format
         """
         index = 0
+        match_found = False
         for a, b in zip(ovs_extra, cmd_map['command']):
             if not re.match(b, a, re.IGNORECASE):
                 return False
@@ -1727,11 +1728,13 @@ class NmstateNetConfig(os_net_config.NetConfig):
                     if cfg["nm_config"]:
                         key = cfg["nm_config"]
                         config[key] = value
+                        match_found = True
                     elif cfg['nm_config_regex']:
                         m = re.search(cfg['nm_config_regex'], ovs_extra[index])
                         if m:
                             key = m.group(1)
                             config[key] = value
+                            match_found = True
                         else:
                             msg = (
                                 "ovs_extra: Invalid format detected.\n"
@@ -1744,6 +1747,7 @@ class NmstateNetConfig(os_net_config.NetConfig):
                     logger.info(
                         "%s=%s", "->".join(cfg["sub_tree"] + [key]), value
                     )
+        return match_found
 
     def get_handled_ovs_extra(self, original_list, unhandled_list):
         """Get ovs_extra commands that were successfully handled.
@@ -1867,15 +1871,37 @@ class NmstateNetConfig(os_net_config.NetConfig):
         # ovs-vsctl br-set-external-id $name key [value]
         # ovs-vsctl del-controller $name
 
+        unhandled_list = []
+
         for ovs_extra in ovs_extras:
             logger.info("%s: Parse - %s", name, ovs_extra)
             ovs_extra_cmd = ovs_extra.split(' ')
+
+            # Initialize unhandled flag to true before the for loop
+            unhandled_flag = True
+
             for cmd_map in cfg_eq_val_pair:
-                # TODO(arn): Handle return value (match_count,
-                # mismatch_count) for better error handling
-                self._ovs_extra_cfg_eq_val(ovs_extra_cmd, cmd_map, data)
-            for cmd_map in cfg_val_pair:
-                self._ovs_extra_cfg_val(ovs_extra_cmd, cmd_map, data)
+                mc, msc = self._ovs_extra_cfg_eq_val(
+                    ovs_extra_cmd, cmd_map, data)
+                # If no mismatches and there are matches, command was handled
+                if msc == 0 and mc > 0:
+                    unhandled_flag = False
+
+            # If unhandled_flag is true, execute the next for loop
+            if unhandled_flag:
+                for cmd_map in cfg_val_pair:
+                    match_found = self._ovs_extra_cfg_val(
+                        ovs_extra_cmd, cmd_map, data)
+                    # If match_found is set to true, unhandled_flag = false
+                    if match_found:
+                        unhandled_flag = False
+
+            # Check unhandled_flag and add to unhandled list
+            if unhandled_flag:
+                unhandled_list.append(ovs_extra)
+
+        # Just return unhandled list
+        return unhandled_list
 
     def parse_ovs_extra_for_dpdk(self, ovs_extras, dpdk_port, data):
         """Parse ovs extra for bridges, ports
@@ -1942,8 +1968,8 @@ class NmstateNetConfig(os_net_config.NetConfig):
             logger.info("%s: Parse - %s", iface_name, ovs_extra)
             ovs_extra_cmd = ovs_extra.split(' ')
 
-            # Initialize error flag to true before the for loop
-            error_flag = True
+            # Initialize unhandled flag to true before the for loop
+            unhandled_flag = True
 
             for cmd_map in cfg_eq_val_pair:
                 mc, msc = self._ovs_extra_cfg_eq_val(ovs_extra_cmd, cmd_map,
@@ -1951,10 +1977,10 @@ class NmstateNetConfig(os_net_config.NetConfig):
                 # If no mismatches and there are matches, command was
                 # handled successfully
                 if msc == 0 and mc > 0:
-                    error_flag = False
+                    unhandled_flag = False
 
-            # Check error_flag and add to unhandled list
-            if error_flag:
+            # Check unhandled_flag and add to unhandled list
+            if unhandled_flag:
                 unhandled_list.append(ovs_extra)
 
         # Return unhandled list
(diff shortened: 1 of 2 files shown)


os_net_config/impl_nmstate.py
```diff
@@ -2043,6 +2085,27 @@ def add_bridge(self, bridge, dpdk=False):
         else:
             mac = None
 
+        # Chain ovs_extra parsing to categorize commands by type
+        dummy_data = {}
+        unhandled_iface_list = self.parse_ovs_extra_for_iface(
```
karthiksundaravel, line 2090: May be `unhandled_ovs_extra_iface` sounds better than unhandled_iface_list ?

os_net_config/impl_nmstate.py
```diff
@@ -2043,6 +2085,27 @@ def add_bridge(self, bridge, dpdk=False):
...
+            bridge.ovs_extra, bridge.name, dummy_data)
+        interface_ovs_extra = self.get_handled_ovs_extra(
+            bridge.ovs_extra, unhandled_iface_list)
+
+        unhandled_port_list = self.parse_ovs_extra_for_ports(
```
karthiksundaravel, line 2095: May be unhandled_ovs_extra_port sounds better than unhandled_port_list ?

os_net_config/impl_nmstate.py
```diff
@@ -2043,6 +2085,27 @@ def add_bridge(self, bridge, dpdk=False):
...
+            unhandled_iface_list, bridge.name, dummy_data)
+        port_ovs_extra = self.get_handled_ovs_extra(
+            unhandled_iface_list, unhandled_port_list)
+
+        unhandled_bridge_list = self.parse_ovs_extra_for_bridge(
```
karthiksundaravel, line 2100: May be unhandled_ovs_extra sounds better than unhandled_bridge_list ?

os_net_config/impl_nmstate.py
```diff
@@ -2043,6 +2085,27 @@ def add_bridge(self, bridge, dpdk=False):
...
+            unhandled_port_list, bridge.name, dummy_data)
+
+        # Check for completely unhandled commands
+        if unhandled_bridge_list:
```
karthiksundaravel, line 2104: if `unhandled_ovs_extra`:

os_net_config/impl_nmstate.py
```diff
@@ -2043,6 +2085,27 @@ def add_bridge(self, bridge, dpdk=False):
...
+            msg = (f"{bridge.name}: Unhandled ovs_extra commands: "
+                   f"{unhandled_bridge_list}. Check the format")
+            raise os_net_config.ConfigurationError(msg)
+
```
karthiksundaravel, line 2108: bridge_ovs_extra = self.get_handled_ovs_extra(
            unhandled_ovs_extra_port, unhandled_ovs_extra)

os_net_config/impl_nmstate.py
```diff
@@ -2052,13 +2115,14 @@ def add_bridge(self, bridge, dpdk=False):
...
+            domain=bridge.domain, hwaddr=mac, ovs_extra=interface_ovs_extra)
         self.add_ovs_interface(ovs_interface_port)
 
         ovs_int_port = {'name': ovs_interface_port.name}
-        if bridge.ovs_extra:
-            self.parse_ovs_extra_for_ports(bridge.ovs_extra,
-                                           bridge.name, ovs_int_port)
+        ports_unhandled_list = []
+        if port_ovs_extra:
+            ports_unhandled_list = self.parse_ovs_extra_for_ports(
```
karthiksundaravel, line 2124: It is not necessary to re-compute the applicable ovs_extra for ports and bridges again. Please use port_ovs_extra and bridge_ovs_extra computed earlier

os_net_config/impl_nmstate.py
```diff
@@ -2099,13 +2163,26 @@ def add_bridge(self, bridge, dpdk=False):
                            OvsDB.OTHER_CONFIG: {}}
         data[OVSBridge.CONFIG_SUBTREE][
             OVSBridge.ALLOW_EXTRA_PATCH_PORTS] = True
-        bridge.ovs_extra.append("set bridge %s other-config:mac-table-size=%d"
-                                % (bridge.name, common.MAC_TABLE_SIZE))
+        # Combine extracted bridge commands with auto-generated commands
+        actual_bridge_commands = self.get_handled_ovs_extra(
```
karthiksundaravel, line 2167: It would be computed already in bridge_ovs_extra

os_net_config/impl_nmstate.py
```diff
@@ -2099,13 +2163,26 @@ def add_bridge(self, bridge, dpdk=False):
...
+        actual_bridge_commands.extend(ports_unhandled_list)
+        actual_bridge_commands.append(
+            "set bridge %s other-config:mac-table-size=%d"
+            % (bridge.name, common.MAC_TABLE_SIZE))
         if bridge.primary_interface_name:
             mac = self.interface_mac(bridge.primary_interface_name)
-            bridge.ovs_extra.append("set bridge %s other_config:hwaddr=%s" %
-                                    (bridge.name, mac))
-        self.parse_ovs_extra_for_bridge(bridge.ovs_extra, bridge.name, data)
+            actual_bridge_commands.append(
```
karthiksundaravel, line 2175: bridge_ovs_extra.append() shall be used instead

os_net_config/impl_nmstate.py
```diff
@@ -2276,8 +2353,12 @@ def add_ovs_interface(self, ovs_interface):
 
         # Parse ovs_extra commands for the interface
         if ovs_interface.ovs_extra:
-            self.parse_ovs_extra_for_iface(
```
karthiksundaravel, line 2279: unhandled_ovs_extra instead of unhandled_list

karthiksundaravel (changes requested):
Changes look good, I've suggested some variable names instead.

karthiksundaravel:
@dsneddon We know using ifcfg provider all the options supported by ovs could be configured with OVS_EXTRA field. However with nmstate provider all possible options supported by ovs could not be configured, the idea is to identify invalid format or unhandled ovs_extra passed to os-net-config and raise an exception to notify the same.

os_net_config/impl_nmstate.py
```diff
@@ -1867,15 +1871,37 @@ def parse_ovs_extra_for_bridge(self, ovs_extras, name, data):
...
         # ovs-vsctl del-controller $name
 
+        unhandled_list = []
+
         for ovs_extra in ovs_extras:
             logger.info("%s: Parse - %s", name, ovs_extra)
             ovs_extra_cmd = ovs_extra.split(' ')
+
+            # Initialize error flag to true before the for loop
+            error_flag = True
```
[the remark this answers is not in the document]
abhiramnarayana: Sure. Will change it to unhandled_flag. Also will do in other functions also as we have similar changes in couple of other places as well.

os_net_config/impl_nmstate.py
abhiramnarayana: Sure. Will change it.

os_net_config/impl_nmstate.py
abhiramnarayana: Sure.

os_net_config/impl_nmstate.py
abhiramnarayana: Sure.

os_net_config/impl_nmstate.py
abhiramnarayana: Note this call is need because the earlier pass was with "dummy_data" and this one is with "ovs_int_port" . So this will be needed I feel.

abhiramnarayana: As discussed, the call is needed but reduce usage of 1 variable list i.e. ports_unhandled_list as that is already calculated.

karthiksundaravel:
> @karthiksundaravel / @abhiramnarayana , do we support ovs_extra/_options for vlan interface (which will be part of ovs_bridge) ?

https://github.com/os-net-config/os-net-config/blob/master/os_net_config/impl_nmstate.py#L1993C9-L1993C34

dsneddon:
What about dpdk-lsc-interrupt option for DPDK ports? I know that some customers use this setting when configuring DPDK ports for bonds:

https://docs.openvswitch.org/en/latest/topics/dpdk/phy/#link-state-change-lsc-detection-configuration

In response to:
> hi @vcandapp , As discussed vlan ovs_options are not supported currently for nmstate provider. I have raised a JIRA so that we can track it separately ([issues.redhat.com/browse/OSPRH-19061](https://issues.redhat.com/browse/OSPRH-19061)) Anyways, this PR will not change anything existing or already supported.

dsneddon:
I have some concerns about this, and I wonder if we could possibly request better error reporting from the nmnstate Python library so that we were not doing the checking in os-net-config. What happens when nmstate adds support for new options? We will need to monitor their development closely and add those options to os-net-config. However if we were to pass any option and then report a meaningful error from nmstate if the option didn't work then if support for a new option were added it would be usable right away.

Is it worthwhile to create a feature request for nmstate library to specifically report on unhandled OVS options so that the checking is done in the nmstate schema validation instead of here in os-net-config?

dsneddon (commented):
I have a couple of comments inline.

os_net_config/impl_nmstate.py
```diff
@@ -1744,6 +1747,7 @@ def _ovs_extra_cfg_val(self, ovs_extra, cmd_map, data):
                     logger.info(
                         "%s=%s", "->".join(cfg["sub_tree"] + [key]), value
                     )
+        return match_found
```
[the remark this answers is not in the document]
abhiramnarayana: Yeah, At a first look it appear so. But believe me I have tried with many many config files!. At times, I feel have we started on a path which we shouldnt have even started!.. i.e. Raising error on wrong configs.. But since we have already here we are, kindly see my below explanation.

The top level return False handles if the command valid and does early exit if it is invalid. And no need to enter further.
Example: invalid-command br-ex key value

But, a supported command but broken or half command will still creep in
This return match_found can be still False in such cases. (missing key value)
Example: br-set-external-id br-ex

The correct command will return match_found as True 
Example: br-set-external-id br-ex key value

Hence, I feel better we keep this for now as this is needed. Hope it is fine by you.

PR #269, opened 2025-08-12
utils: refactor DPDK detach with retry mechanism and error handler

karthiksundaravel:
* Add retry logic with 2-second delays for transient failures
* Simplify return codes to 0 (success) or 1 (error)
* Handle both stderr and ProcessExecutionError paths consistently
* Add comprehensive test coverage for all error scenarios


dsneddon (approved):
LGTM

PR #275, opened 2025-08-19
nmstate: keep_addr_on_down dispatcher script

dsneddon:
This patch adds dispatcher script support for interfaces with static IPv6 addresses which sets the sysctl for keep_addr_on_down in order to have static IPv6 addresses act similar to static IPv4 addresses (interfaces keep their IP address when down or on reboot without requiring automatic configuration).

A follow-up patch will attempt to replicate this behavior for ifcfg files that have NM_CONTROLLED=yes, but that will require handling the dispatcher script fully within os-net-config without the benefit of native support for dispatcher script provided by nmstate.

### DIFF
--- a/os_net_config/impl_nmstate.py
+++ b/os_net_config/impl_nmstate.py
@@ -86,6 +86,8 @@ done
 
 ETHTOOL_SCRIPT = "{ethtool_cmd} {ethtool_opts} #ETHTOOL"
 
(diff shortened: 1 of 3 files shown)


dsneddon:
recheck

dsneddon:
recheck

karthiksundaravel (approved):
LGTM. Thanks.

PR #283, opened 2025-08-31
utils: add write_bonding_masters function for Linux bond management

abhiramnarayana:
This patch adds write_bonding_masters() utility function to manage Linux bond interfaces by writing to /sys/class/net/bonding_masters. The function supports both creating and removing bonds through the bonding_masters sysfs interface.

The function will be used by device removal functionality to properly clean up Linux bond interfaces from the kernel bonding subsystem.


karthiksundaravel (commented):
One minor comment. Otherwise looks good to me

PR #286, opened 2025-09-04
nmstate: add vlan over bond as ports to ovs bridges

karthiksundaravel:


dsneddon:
I would like to see a commit message description before this merges (assuming it graduates from DNM to a real pull request).

os_net_config/impl_nmstate.py
```diff
@@ -954,7 +954,7 @@ def interface_mac(self, iface):
         if iface_data and Interface.MAC in iface_data:
             return iface_data[Interface.MAC]
 
-    def get_ovs_ports(self, members):
+    def get_ovs_ports(self, bridge_name, members):
         """Get the ovs ports in nmstate schema format
 
         :param members: The members of the ovs/ovs-user bridges
```
vcandapp, line 960: New param bridge_name needs to be updated

os_net_config/impl_nmstate.py
```diff
@@ -976,13 +976,16 @@ def get_ovs_ports(self, members):
             # like eth0.605
             elif re.match(r'\w+\.\d+$', member):
                 vlan_id = int(member.split('.')[1])
+                vlan_dev = member.split('.')[0]
```
vcandapp, line 979: In the if/elif block - only the vlan_id extraction differs. Can the schema initialization be kept same for bond_api.143 and vlan143 ?

os_net_config/impl_nmstate.py
karthiksundaravel: Done

vcandapp (approved):
Changes look good, thanks

PR #291, opened 2025-09-16
doc: update the doc for the exit codes

karthiksundaravel:
The usage.rst shall be modified to indicate the exit code when detailed exit codes are enabled


karthiksundaravel:
recheck

PR #295, opened 2025-09-22
cli: schema validation is needed only for user input

vcandapp:
In get_iface_config() should do the schema validation as-is on the YAML data in user input. Schema checks for internal types like nic_mapping, persist_mapping  is not needed


karthiksundaravel (approved):
lgtm. Thanks

PR #299, opened 2025-09-24
ifcfg: set sysctl to set keep_addr_on_down

dsneddon:
Write a sysctl.d config file in /usr/lib/sysctl.d when
IPv6 addresses are used to set keep_addr_on_down as a
default and for all existing interfaces. This behavior
is default for IPv4 addresses and needs to be used for
IPv6 addresses on systems that are routing IPv6.

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -72,6 +72,12 @@ SRIOV_CONFIG_FILE = '/var/lib/os-net-config/sriov_config.yaml'
 
 DCB_CONFIG_FILE = '/var/lib/os-net-config/dcb_config.yaml'
 
+# Directory to write sysctl settings to
+# Format of the file shall be
+# net.ipv6.conf.default.keep_addr_on_down=1
+# net.ipv6.conf.all.keep_addr_on_down=1
+SYSCTLD_CONF_DIR = '/usr/lib/sysctl.d'
+
 _SYS_BUS_PCI_DEV = '/sys/bus/pci/devices'
 SYS_CLASS_NET = '/sys/class/net'
 _LOG_FILE = '/var/log/os-net-config.log'
@@ -190,6 +196,22 @@ def write_yaml_config(filepath, data):
     os.makedirs(os.path.dirname(filepath), exist_ok=True)
     with open(filepath, 'w') as f:
         yaml.safe_dump(data, f, default_flow_style=False)
+
+
+def write_sysctld_conf(filename, sysctls=[]):
+    filepath = SYSCTLD_CONF_DIR + '/' + filename
+    filedata = '# Written by os-net-config, do not modify\n'
+    filedata += '\n'.join(sysctls) + '\n'
+    logger.info("Writing file %s with content %s", filepath, filedata)
+    if get_noop():
+        return
+    try:
+        with open(filepath, 'w') as f:
+            f.write(str(filedata))
+            return True
+    except IOError:
+        logger.error("Error writing file: %s", filename)
+        raise
 
 
 def update_dcb_map(device, pci_addr, driver, noop, dscp2prio=None):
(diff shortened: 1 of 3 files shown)


os_net_config/impl_ifcfg.py
```diff
@@ -750,6 +750,11 @@ def _add_common(self, base_opt):
             data += "MTU=%i\n" % base_opt.mtu
         if base_opt.use_dhcpv6 or base_opt.v6_addresses():
             data += "IPV6INIT=yes\n"
+            keep_addr_sysctls = ["net.ipv6.conf.default.keep_addr_on_down=1",
+                                 "net.ipv6.conf.all.keep_addr_on_down=1"]
+            common.write_sysctld_conf("90-v6_keep_addr.conf",
+                                      keep_addr_sysctls)
+            utils.set_keep_addr_sysctl(base_opt.name, self.noop)
```
dsneddon, line 757: Note that when booting setting net.ipv6.conf.default and net.ipv6.conf.all will ensure that all interfaces get keep_addr_on_down set as well. However while running we must set the value on the interface directly.

os_net_config/impl_ifcfg.py
```diff
@@ -750,6 +750,9 @@ def _add_common(self, base_opt):
             data += "MTU=%i\n" % base_opt.mtu
         if base_opt.use_dhcpv6 or base_opt.v6_addresses():
             data += "IPV6INIT=yes\n"
+            keep_addr_sysctl = "net.ipv6.conf.default.keep_addr_on_down=1\n" +\
+                               "net.ipv6.conf.all.keep_addr_on_down=1"
+            common.write_sysctld_conf("90-v6_keep_addr.conf", keep_addr_sysctl)
```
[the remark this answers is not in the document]
dsneddon: It turns out that the conf.all and conf.default are needed on boot, but while running the interface must be set directly. The new patchset sets the value on the interface directly and then the file that sets all and default will ensure that all interfaces get set on reboot.

os_net_config/common.py
```diff
@@ -192,6 +198,22 @@ def write_yaml_config(filepath, data):
...
+    filepath = SYSCTLD_CONF_DIR + '/' + filename
+    filedata = '# Written by os-net-config, do not modify\n'
+    filedata += '\n'.join(sysctls) + '\n'
+    logger.info("Writing file %s with content %s", filepath, filedata)
+    if get_noop():
+        return
+    try:
+        with open(filepath, 'w') as f:
+            f.write(str(filedata))
+            return True
```
vcandapp, line 211: @dsneddon / @karthiksundaravel  , currently in the " ifcfg_requires_restart()" - we have IP addr in permitted changes. But, for v6 - does IPV6_SET_SYSCTLS -> yes / no requires a restart ?

os_net_config/impl_ifcfg.py
```diff
@@ -750,6 +750,11 @@ def _add_common(self, base_opt):
             data += "MTU=%i\n" % base_opt.mtu
         if base_opt.use_dhcpv6 or base_opt.v6_addresses():
             data += "IPV6INIT=yes\n"
+            keep_addr_sysctls = ["net.ipv6.conf.default.keep_addr_on_down=1",
+                                 "net.ipv6.conf.all.keep_addr_on_down=1"]
+            common.write_sysctld_conf("90-v6_keep_addr.conf",
```
vcandapp, line 755: @karthiksundaravel - in purge provider - this file should be addes right ?

os_net_config/impl_ifcfg.py
dsneddon: The v6 routes are also retained when the addresses are retained, so routes come up faster with this in place. Does thst answer your question?

PR #300, opened 2025-09-27
nmstate: fix sysctl path for tests

karthiksundaravel:
The sysctl_path() needs to be stubbed for the unit tests to pass.


vcandapp:
recheck

PR #301, opened 2025-09-30
nmstate: fix unit tests from accessing nmstate api

karthiksundaravel:
The unit tests shall not use the api's exposed by nmstate.


vcandapp:
LGTM

PR #302, opened 2025-10-04
Group config

karthiksundaravel:


os_net_config/cli.py
```diff
@@ -271,24 +275,28 @@ def main(argv=sys.argv, main_logger=None):
...
+                persist_mapping,
+                strict_validate=opts.exit_on_validation_errors,
+            )
+        except objects.InvalidConfigException as e:
+            main_logger.error(
+                "%s: Schema validation failed with error: \n%s", section, e
+            )
+            return get_exit_code(
+                opts.detailed_exit_codes,
+                onc_ret_code | ExitCode.REMOVE_CONFIG_FAILED
```
abhiramnarayana: hi @karthiksundaravel ,
Here do we need to set Exit code based on the section ? It says only REMOVE_CONFIG_FAILED ?

os_net_config/cli.py
```diff
@@ -298,39 +306,6 @@ def main(argv=sys.argv, main_logger=None):
...
-            strict_validate=opts.exit_on_validation_errors,
-        )
-    except objects.InvalidConfigException as e:
-        main_logger.error(
-            "Schema validation failed for network_config with error: \n%s", e
-        )
-        return get_exit_code(opts.detailed_exit_codes,
-                             onc_ret_code | ExitCode.SCHEMA_VALIDATION_FAILED)
-
-    if not iface_array:
```
abhiramnarayana: hi @karthiksundaravel ,
If network_config section was not there we used to fail before right? Now, this kind of check not needed for the network_config section? remove_config its not mandatory. But what about network_config section?

karthiksundaravel: Added now :)

PR #307, opened 2025-10-13
cli: add --remove-config flag to enable parsing this section

vcandapp:
- Adding new flag '--remove-config', to enable conditional parsing of this section in config YAML
- It is disabled by default
- Added test case for same

### DIFF
--- a/os_net_config/cli.py
+++ b/os_net_config/cli.py
@@ -180,6 +180,15 @@ def parse_opts(argv):
              "(WARNING, permanently renames nics).",
         required=False)
 
+    parser.add_argument(
+        '--remove-config',
+        dest="remove_config",
+        action='store_true',
+        help="""Apply remove_config section before applying network_config."""
+        """ This is useful to clean previously configured interfaces. """
+        """Disabled by default.""",
+        required=False)
+
     opts = parser.parse_args(argv[1:])
 
     return opts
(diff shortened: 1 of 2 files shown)


karthiksundaravel (dismissed):
LGTM. Thanks

os_net_config/cli.py
```diff
@@ -339,17 +348,33 @@ def main(argv=sys.argv, main_logger=None):
...
     if config_data["remove_config"]:
-        ret_code = apply_remove_config(
-            config_data["remove_config"], opts.root_dir, opts.noop
-        )
-        if ret_code == ExitCode.REMOVE_CONFIG_FAILED:
-            main_logger.error("Failed to apply remove_config")
-            return get_exit_code(
-                opts.detailed_exit_codes,
-                onc_ret_code | ExitCode.REMOVE_CONFIG_FAILED
+        if opts.remove_config:
```
karthiksundaravel: can we keep opts.remove_config in the outer "if", which will align with the other cli args

os_net_config/cli.py
```diff
@@ -339,17 +348,33 @@ def main(argv=sys.argv, main_logger=None):
...
+                main_logger.error("Failed to apply remove_config")
+                return get_exit_code(
+                    opts.detailed_exit_codes,
+                    onc_ret_code | ExitCode.REMOVE_CONFIG_FAILED
+                )
+            else:
+                main_logger.info("remove_config applied successfully")
         else:
-            main_logger.info("remove_config applied successfully")
+            onc_ret_code |= ExitCode.REMOVE_CONFIG_FAILED
```
karthiksundaravel: When --remove-config is not added, the config_data["remove_config"] need not be verified

karthiksundaravel (changes requested):

os_net_config/cli.py
vcandapp: Done

os_net_config/cli.py
vcandapp: Done

karthiksundaravel (changes requested):

PR #312, opened 2025-10-27
ifcfg: report failure in bringing up ovs bridges

karthiksundaravel:
The ifup scripts fail to return non-zero error codes for OVS devices, but the error messages are available in stderr and stdout. These error messages are used to raise exceptions and handle the failure of ifup scripts.

### DIFF
--- a/os_net_config/__init__.py
+++ b/os_net_config/__init__.py
@@ -490,14 +490,34 @@ class NetConfig(object):
 
         Print a message and run a command with processutils
         in noop mode, this just prints a message.
+
+        :param parse_errors: If True, raises ProcessExecutionError when 'error'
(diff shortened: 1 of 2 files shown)


os_net_config/__init__.py
```diff
@@ -490,14 +490,34 @@ def execute(self, msg, cmd, *args, **kwargs):
 
         Print a message and run a command with processutils
         in noop mode, this just prints a message.
+        
+        :param parse_errors: If True, raises ProcessExecutionError when 'error'
```
vcandapp: The synopsis needs update (parse_errors is not param)

abhiramnarayana:
recheck

PR #313, opened 2025-10-27
ifcfg: add ovs_extra for linux bonds

karthiksundaravel:
This commit adds support for ovs_extra configuration in Linux bonds,
allowing users to specify additional OVS commands for bonded interfaces
when used as OVS bridge members.

Changes:
- Add ovs_extra parameter to LinuxBond constructor and from_json method
- Process ovs_extra templates using format_ovs_extra function
- Validate that ovs_extra is only allowed for OVS bridge members
- Extend bridge's ovs_extra list with bond's ovs_extra commands
- Update schema to include ovs_extra property for linux_bond type


karthiksundaravel:
recheck

abhiramnarayana (approved):
LGTM.

PR #314, opened 2025-10-31
vlan: check if wrong device param is set for VLAN member

vcandapp:
- Added check for ovs_bridge members, when user configures a wrong "device" param for bridge -> bond -> VLAN member use case
- Not funtionality changes, just log an error message

Fixes: https://issues.redhat.com/browse/OSPRH-21391

### DIFF
--- a/os_net_config/objects.py
+++ b/os_net_config/objects.py
@@ -749,6 +749,19 @@ class OvsBridge(_BaseOpts):
             member.bridge_name = name
             if isinstance(member, SriovVF):
                 OvsBridge.update_vf_config(member)
+            if isinstance(member, Vlan):
+                base_iface = getattr(member, 'device', None)
+                if base_iface and base_iface != name:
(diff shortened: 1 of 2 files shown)


karthiksundaravel (commented):
I think in class Vlan, a check could be added when vlan_dev.linux_bond_name exists or vlan_dev.bridge_name exists, the device field should match one of these.

os_net_config/objects.py
```diff
@@ -907,6 +920,19 @@ def __init__(self, name, use_dhcp=False, use_dhcpv6=False, addresses=None,
         for member in self.members:
             member.linux_bridge_name = name
             member.ovs_port = False
+            if isinstance(member, Vlan):
```
karthiksundaravel: same needs to be added for linux_bond as well

karthiksundaravel (approved):

PR #327, opened 2025-11-26
nmstate: delete vlan devices with name <iface>.<vlanid>

karthiksundaravel:
The vlan device could be provided in the form <interface>.<vlan-id>.
This format of representation needs to be handled during the destroy of
vlans.


abhiramnarayana (approved):
LGTM

karthiksundaravel:
recheck

PR #331, opened 2025-12-08
ifcfg: add linux bond mode validation and tests

vcandapp:
- get_type_value() method moved to common.pu
- added new get_bonding_options() method for ifcfg provider
- Validate bond modes against linux_bond_mode dictionary
- Add test cases for valid/invalid bond modes

Depends-On: https://github.com/os-net-config/os-net-config/pull/332

### DIFF
--- a/os_net_config/common.py
+++ b/os_net_config/common.py
@@ -149,6 +149,22 @@ def get_timestamp():
     return time.strftime("%Y%m%d_%H%M%S", time.localtime())
 
 
+def get_type_value(str_val):
+    """Convert string value to appropriate type (int, bool, or str).
+
+    :param str_val: String value to convert
+    :returns: Converted value (int, bool, or str)
+    """
+    if isinstance(str_val, str):
+        if str_val.isdigit():
+            return int(str_val)
+        if str_val.lower() in ['true', 'yes', 'on']:
+            return True
+        if str_val.lower() in ['false', 'no', 'off']:
+            return False
+    return str_val
+
+
 def get_dev_path(ifname, path=None):
     if not path:
         path = ""
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -159,6 +159,24 @@ def stop_dhclient_process(interface):
             )
 
 
+def get_bonding_options(bond_options_str):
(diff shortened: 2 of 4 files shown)


karthiksundaravel (changes requested):
The parse functions are added in impl_nmstate.py are specific to nmstate scheme of things. If ifcfg needs the same, you can define something similar in impl_ifcfg.py instead of moving it to common.py.  common.py usually has the driver or low level handling stuffs and not about parsing the user input

os_net_config/common.py
```diff
@@ -149,6 +149,22 @@ def get_timestamp():
     return time.strftime("%Y%m%d_%H%M%S", time.localtime())
 
 
+def _get_type_value(str_val):
```
abhiramnarayana: Earlier this was in a file and only used there and it was private api. Now, since its common underscore not needed. Right?

os_net_config/impl_ifcfg.py
```diff
@@ -215,6 +233,25 @@ def __init__(self, noop=False, root_dir=''):
             "sriov_pf": [],
             "dpdk_port": [],  # PCI address of the DPDK port is added here
         }
+        # Linux bond mode mapping
```
karthiksundaravel: keep this local to the parsing function

os_net_config/impl_ifcfg.py
```diff
@@ -688,6 +725,27 @@ def _add_common(self, base_opt):
                 members = [member.name for member in base_opt.members]
                 self.member_names[base_opt.name] = members
             if base_opt.bonding_options:
+                # Validate bond mode against dict before writing to ifcfg file
+                bond_opts = get_bonding_options(
+                    base_opt.bonding_options)
+                if 'mode' in bond_opts:
+                    mode_value = bond_opts['mode']
```
karthiksundaravel: if mode_value not in self.linux_bond_mode.keys():
   raise ValueError(...)

karthiksundaravel (changes requested):

os_net_config/impl_ifcfg.py
vcandapp: Done

karthiksundaravel (approved):

PR #332, opened 2025-12-10
Change parent job for content provider

danpawlik:
The old parent does not contain required variables, so later jobs that relay on it are failing.


karthiksundaravel:
Please keep the title like
ci: change parent job for content provider

PR #335, opened 2026-02-19
ifcfg: update accept_ra when no static addr is specified

vcandapp:
This change adds proper IPv6 RA handling for interfaces without static addresses, preventing unintended IPv6 autoconfiguration based on global sysctl settings.

- If IPv6 is enabled globally AND RA is disabled by default, update ifcfg- to disable auto-conf
- Added new methods to read sysctl
- Updated tests cases for few possible scenarios

### DIFF
--- a/os_net_config/impl_ifcfg.py
+++ b/os_net_config/impl_ifcfg.py
@@ -784,6 +784,21 @@ class IfcfgNetConfig(os_net_config.NetConfig):
                     if ":" in route.next_hop:
                         data += f"IPV6_DEFAULTGW={route.next_hop}\n"
                         data += f"IPV6_DEFAULTDEV={base_opt.name}\n"
+        # If no static addr is configured and IPv6 is globally enabled with
+        # accept_ra disabled, explicitly disable RA to avoid unintended
+        # IPv6 auto-configuration.
+        # Also make sure accept_ra is not set for the "dynamic IPv4 only"
+        # (for backward compatibility with existing ifcfg-* behavior)
+        elif not base_opt.use_dhcp:
+            ipv6_disabled = utils.get_sysctl_value(
+                'net.ipv6.conf.default.disable_ipv6')
+            accept_ra_default = utils.get_sysctl_value(
+                'net.ipv6.conf.default.accept_ra')
+
+            if (ipv6_disabled == '0' and accept_ra_default == '0'):
+                data += "IPV6_AUTOCONF=no\n"
+                data += "IPV6_SET_SYSCTLS=yes\n"
+                data += "IPV6_FORCE_ACCEPT_RA=no\n"
 
         if base_opt.hwaddr:
             data += "HWADDR=%s\n" % base_opt.hwaddr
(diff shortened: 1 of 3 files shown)


os_net_config/impl_ifcfg.py
```diff
@@ -784,6 +784,18 @@ def _add_common(self, base_opt):
                     if ":" in route.next_hop:
                         data += f"IPV6_DEFAULTGW={route.next_hop}\n"
                         data += f"IPV6_DEFAULTDEV={base_opt.name}\n"
+        elif not base_opt.addresses:
+            # If i/f has no static addresses but IPv6 is enabled globally and
+            # accept_ra is disabled, disable RA to prevent unwanted auto-conf
+            ipv6_disabled = utils.get_sysctl_value(
```
abhiramnarayana: hi @vcandapp , 
Now that we are calling this new util function in this path. In the new tests which you have added already I see you have added the stub
Stubs are needed in some existing unit tests also I think where this gets called right? Can you check  examples like test_add_base_interface, test_add_vlan_mtu_1500

os_net_config/impl_ifcfg.py
```diff
@@ -784,6 +784,18 @@ def _add_common(self, base_opt):
                     if ":" in route.next_hop:
                         data += f"IPV6_DEFAULTGW={route.next_hop}\n"
                         data += f"IPV6_DEFAULTDEV={base_opt.name}\n"
+        elif not base_opt.addresses:
```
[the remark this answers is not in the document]
vcandapp: Yes, this is a special case when no dhcp6 and no static ipv6 is configured

os_net_config/impl_ifcfg.py
abhiramnarayana: @karthiksundaravel .. any thoughts?

karthiksundaravel (changes requested):

os_net_config/impl_ifcfg.py
vcandapp: @abhiramnarayana / @karthiksundaravel - there is no error reported in reading from sysctl - just debug log and None

os_net_config/impl_ifcfg.py
```diff
@@ -784,6 +784,18 @@ def _add_common(self, base_opt):
                     if ":" in route.next_hop:
                         data += f"IPV6_DEFAULTGW={route.next_hop}\n"
                         data += f"IPV6_DEFAULTDEV={base_opt.name}\n"
+        elif (not base_opt.addresses and not base_opt.use_dhcp):
```
karthiksundaravel: should we use else instead ?

Anyway `not base_opt.addresses` is redundant here as it would be always `true`

From the below table,  the necessary and sufficient check for handling case 1,3 are 
`no use_dhcpv6` and `no base_opt.addresses` which are taken care by `else` 

	use_dhcp	use_dhcpv6	base_opt.addresses    Description
1	False	        False	        False                 NO IPv6 or IPv4 
2	False	        True	        False                 DHCPv6 only + no ipv4
3	True	        False	        False                 Dynamic IPv4 only  
4	True	        True	        False                 (Dual stack) Dynamic IPv6 and v4

karthiksundaravel (changes requested):

karthiksundaravel (approved):
lgtm. Thanks for changing the comments.

karthiksundaravel (approved):

PR #341, opened 2026-04-28
ifcfg: remove IPV6_DEFAULT{GW,DEV} generation

yushoyamaguchi:
# Motivation
`IPV6_DEFAULTGW` and `IPV6_DEFAULTDEV` in a per-interface `ifcfg` file have no effect on NetworkManager beyond creating a duplicate default route.

NetworkManager reads IPV6_DEFAULTGW from the interface's `ifcfg` to set the gateway on `NMSettingIP6Config`, which generates an additional `::/0` route. Since the default route is already written to `route6-<if>` with the correct metric, this results in two `::/0` entries competing in the FIB when their metrics differ.

IPV6_DEFAULTDEV is never read from a per-interface ifcfg at all; NetworkManager only reads it from /etc/sysconfig/network.

Remove both keys as they are redundant and can cause duplicate default routes.

# The case we are having trouble with.
Our Config YAML of os-net-config
```
addresses:
   - ip_netmask: 2001:268:70fb:1b1c::fc12/64
   routes:
   - default: true
     next_hop: 2001:268:70fb:1b1c::1
     route_options: metric 3707764992
```

Result of kernel FIB
```
default via 2001:268:70fb:1b1c::1 dev ens17f0v0 metric 1 pref medium
default via 2001:268:70fb:1b1c::1 dev ens17f0v0 metric 3707764992 pref medium
```

## Why we want to change metric
We want to add a default route which is weaker than BGP route from FRR.
https://docs.frrouting.org/en/latest/zebra.html#administrative-distance


os_net_config/impl_ifcfg.py
```diff
@@ -882,11 +881,6 @@ def _add_common(self, base_opt):
                     secondaries_v6 = " ".join(map(lambda a: a.ip_netmask,
                                                   v6_addresses[1:]))
                     data += "IPV6ADDR_SECONDARIES=\"%s\"\n" % secondaries_v6
-            for route in base_opt.routes:
-                if route.default or (route.ip_netmask == "::/0"):
-                    if ":" in route.next_hop:
-                        data += f"IPV6_DEFAULTGW={route.next_hop}\n"
```
karthiksundaravel, line 888: Looks good. We need to check if this causes any regression due to https://github.com/os-net-config/os-net-config/pull/261

yushoyamaguchi: @karthiksundaravel 
The lines generating `IPV6_DEFAULTGW` and `IPV6_DEFAULTDEV` were indeed touched in https://github.com/os-net-config/os-net-config/pull/261, but that was purely a cosmetic change — rewriting .format() calls as f-strings with no difference in output.
It was incidental to the actual intent of that commit, which was to ensure `IPV6_FORCE_ACCEPT_RA=no` is always written when a static IPv6 address is configured, regardless of defroute.
My patch removes those two lines(about `IPV6_DEFAULTGW` and `IPV6_DEFAULTDEV`) entirely, but the `IPV6_FORCE_ACCEPT_RA` logic introduced by https://github.com/os-net-config/os-net-config/pull/261 is untouched and continues to work as intended.

yushoyamaguchi:
```
(.venv-os-net-config) yamaguchi@yamaguchi-cilium-dev:~/mship-oss/os-net-config$ git branch
  master
* rm_ipv6_default
(.venv-os-net-config) yamaguchi@yamaguchi-cilium-dev:~/mship-oss/os-net-config$ cat ../accept_ra_demo.yaml 
network_config:
  - type: interface
    name: em1
    mtu: 9000
    addresses:
      - ip_netmask: 2001:db8:a::10/64   # Static IPv6
    routes:
      - ip_netmask: ::/0
        next_hop: 2001:db8:a::1         # This part will be written to route6-em1
(.venv-os-net-config) yamaguchi@yamaguchi-cilium-dev:~/mship-oss/os-net-config$ sudo PYTHONPATH="$PWD" /home/yamaguchi/python/.venv-os-net-config/bin/python3     -m os_net_config.cli --noop -c ../accept_ra_demo.yaml -p ifcfg
File: /etc/sysconfig/network-scripts/ifcfg-em1

# This file is autogenerated by os-net-config
DEVICE=em1
ONBOOT=yes
HOTPLUG=no
NM_CONTROLLED=no
PEERDNS=no
MTU=9000
IPV6INIT=yes
IPV6_MTU=9000
IPV6_AUTOCONF=no
IPV6_SET_SYSCTLS=yes
IPV6_FORCE_ACCEPT_RA=no
IPV6ADDR=2001:db8:a::10/64

----
File: /etc/sysconfig/network-scripts/route6-em1

::/0 via 2001:db8:a::1 dev em1

----
```

yushoyamaguchi: @karthiksundaravel 
I've confirmed there are no regression.
`IPV6_FORCE_ACCEPT_RA=no` is written.

yushoyamaguchi: @karthiksundaravel 
Could you review?

karthiksundaravel: @yushoyamaguchi Our team is verifying any regression due to 2 different patches on this. I am waiting for the results before merging.

os_net_config/impl_ifcfg.py
```diff
@@ -349,8 +349,7 @@ def ifcfg_requires_restart(self, filename, new_data):
             'IPADDR', 'NETMASK',
             'MTU', 'ONBOOT', 'ETHTOOL_OPTS',
             'DOMAIN', 'DNS1', 'DNS2',
-            'IPV6_SET_SYSCTLS', 'IPV6_DEFAULTGW',
-            'IPV6_DEFAULTDEV', 'IPV6_FORCE_ACCEPT_RA'
+            'IPV6_SET_SYSCTLS', 'IPV6_FORCE_ACCEPT_RA'
```
vcandapp: This might be needed here to allow users to append these changes to existing config YAML and do a network update without interface restart. @yushoyamaguchi , Can we verify network_update use case also ?

yushoyamaguchi: @vcandapp 
I'll verify soon.
Please let me ask a little.
Is the concern here that existing ifcfg files may still contain `IPV6_DEFAULTGW/IPV6_DEFAULTDEV` (generated by older code), and removing them from `permitted_changes` would cause an unnecessary interface restart when users apply the new config?

vcandapp: @yushoyamaguchi , we need to check the following two behaviour:
- Currenlty even for case where ifcfg doesn't have v6_def_gw and v6_def_dev - user can edit config.yaml to add them - and do network_update (without traffic disruption)
- And yes, case where these param already exist in ifcfg - user does network_update by removing them from config.yaml

vcandapp:
@yushoyamaguchi , CI failed due to merge conflict

os_net_config/impl_ifcfg.py
yushoyamaguchi: @vcandapp 
1. When using this new code, `IPV6_DEFAULTGW`/`IPV6_DEFAULTDEV` will never be written to ifcfg anymore, so there is no network_update case for adding them.
2. For the removal case, I added a test that confirms an unnecessary restart is triggered without the fix, and kept `IPV6_DEFAULTGW`/`IPV6_DEFAULTDEV` in permitted_changes so that removing them from existing ifcfg files does not cause a restart.

yushoyamaguchi:
@vcandapp 
Sorry, I've fixed the conflict.
Please review as you read this(https://github.com/os-net-config/os-net-config/pull/341#discussion_r3394170272).

yushoyamaguchi:
@vcandapp 
The ci-failed label is left over from a previous conflict that has since been resolved. Can you remove it and re-add `ci-gate` to trigger a clean CI run?

yushoyamaguchi:
@vcandapp 
Sorry, could you re-run CI test?

vcandapp (approved):
The updates looks fine to me

yushoyamaguchi:
@vcandapp 
Thank you!
Is it required @karthiksundaravel 's approval for merging?

vcandapp:
> @vcandapp Thank you! Is it required @karthiksundaravel 's approval for merging?

Merged

PR #343, opened 2026-05-08
ifcfg: remove sysctl flag from script

vcandapp:
Setting this flag tirgger a conflict with kernel runtime and the config is unset

### DIFF
--- a/os_net_config/tests/test_impl_ifcfg.py
+++ b/os_net_config/tests/test_impl_ifcfg.py
@@ -785,6 +785,7 @@ class TestIfcfgNetConfig(base.TestCase):
         defaults = {
             'net.ipv6.conf.default.disable_ipv6': '0',
             'net.ipv6.conf.default.accept_ra': '0',
+            'net.ipv6.conf.default.forwarding': '1'
         }
 
         if overrides:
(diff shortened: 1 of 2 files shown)


os_net_config/tests/test_impl_ifcfg.py
```diff
@@ -2419,7 +2419,8 @@ def test_interface_no_ip_with_global_accept_ra_disabled(self):
         self.stub_out(
             'os_net_config.utils.get_sysctl_value',
             self.make_sysctl_stub({
-                'net.ipv6.conf.default.accept_ra': '0'
+                'net.ipv6.conf.default.accept_ra': '0',
+                'net.ipv6.conf.default.forwarding': '0'
```
abhiramnarayana, line 2424: Should this add this in defaults also?
See below line
https://github.com/os-net-config/os-net-config/blob/master/os_net_config/tests/test_impl_ifcfg.py#L824

os_net_config/tests/test_impl_ifcfg.py
abhiramnarayana, line 2416: Should we add a new test as below?
 test_interface_no_ip_with_ipv6_forwarding_enabled that stubs forwarding='1', accept_ra='0', and asserts the output equals _NO_IP (no RA directives).

abhiramnarayana (changes requested):
Have given couple of comments. Kindly check

os_net_config/tests/test_impl_ifcfg.py
vcandapp: Since wallaby is EOL, minimal tests shud be fine, there are other test for _NO_IP case - test_interface_with_addresses_ignores_accept_ra_check, test_interface_no_ip_with_global_accept_ra_enabled

os_net_config/tests/test_impl_ifcfg.py
vcandapp: overrides will handle

abhiramnarayana (approved):

PR #344, opened 2026-05-21
nmstate: add dns_options support for resolver configuration

karthiksundaravel:
Add support for dns_options parameter to configure DNS resolver options like timeout, attempts, and rotate. This allows users to customize resolv.conf behavior through os-net-config.

Changes:
- schema.yaml: Add dns_option_string and list_of_dns_options definitions, add dns_options property to all device types
- objects.py: Add dns_options parameter to _BaseOpts class and base_opts_from_json() method
- impl_nmstate.py: Add _add_dns_options() method, update set_dns() to include options, call dns_options processing from _add_common()

Example usage:
  dns_options:
    - timeout:2
    - attempts:3
    - rotate

Assisted-by: Claude (Anthropic)


os_net_config/schema.yaml
```diff
@@ -131,6 +131,14 @@ definitions:
                 $ref: "#/definitions/domain_name_string"
             minItems: 0
           - $ref: "#/definitions/domain_name_string"
+    dns_option_string:
+        type: string
+        pattern: "^[a-zA-Z][a-zA-Z0-9_-]*(:[a-zA-Z0-9_-]+)?$"
```
vcandapp, line 136: @karthiksundaravel , how do we define the range for each params ? Is it based on the underlying NM library
    **timeout:**
      type: integer
      minimum: 1
      maximum: 30
      default: 5
    **attempts:**
      type: integer
      minimum: 1
      maximum: 5
      default: 2
    **rotate:**
      type: boolean
      default: false

vcandapp:
@karthiksundaravel , shud the doc example - doc/source/config.rst, be updated for DNS section ?

os_net_config/schema.yaml
karthiksundaravel: We don't validate every config given to os-net-config. so IMHO it could be ignored

os_net_config/impl_nmstate.py
```diff
@@ -842,8 +842,13 @@ def set_dns(self):
         :param dns_data:  config json
         :return dns config
         """
-        state = {DNS.KEY: {DNS.CONFIG: {DNS.SERVER: self.dns_data['server'],
-                                        DNS.SEARCH: self.dns_data['domain']}}}
+        dns_config = {
+            DNS.SERVER: self.dns_data['server'],
+            DNS.SEARCH: self.dns_data['domain']
+        }
+        if self.dns_data['options']:
```
vcandapp: Will self.dns_data.get('options') with default [] be safer ?

karthiksundaravel: Will use get()

PR #351, opened 2026-08-05
nmstate: add dpdk-lsc-interrupt option support for DPDK ports

karthiksundaravel:
Add parsing support for the dpdk-lsc-interrupt option in ovs_extra commands for DPDK interfaces. This allows users to configure link state change (LSC) interrupt mode for DPDK ports.
Example usage in ovs_extra:
  - "set interface {name} options:dpdk-lsc-interrupt=true" The option is mapped to the nmstate OVSInterface.Dpdk.LSC_INTERRUPT configuration.


vcandapp (approved):
Changes looks good to me

PR #352, opened 2026-08-10
ifcfg: update order checking for linux_bond and ovs_bond

vcandapp:
The ifcfg_requires_restart check was treating reordered options in BONDING_OPTS and OVS_OPTIONS as a config change, triggering needless interface restarts. Add these keys to the existing sorted comparison alongside BOND_IFACES and OVS_EXTRA.


abhiramnarayana:
recheck

